PR #2, opened 2023-02-08
Added GetAndLock functionality

aemery-cb:


README.md
```diff
@@ -0,0 +1,2 @@
+## Required tools
```
brett19, line 1: This is a great change to make getting started with gocbcorex easier.  However it's not really related to the change at hand.  Would you be able to move this out to an independant commit?

brett19 (changes requested):
One minor change request.  Additionally, could you quickly confirm that the existing GetAndLock implementation in memdx has tests associate with it?

README.md
aemery-cb: I didn't mean to actually open this yet... its still very WIP

PR #3, opened 2023-02-13
Add server error to memdx

chvck:


brett19 (changes requested):

memdx/errors.go
```diff
@@ -34,13 +34,28 @@ func (e protocolError) Unwrap() error {
...
 
+type ServerError struct {
+	Cause          error
+	DispatchedTo   string
+	DispatchedFrom string
+	Opaque         uint32
+}
+
+func (e ServerError) Error() string {
+	return fmt.Sprintf("server error: %s", e.Cause)
```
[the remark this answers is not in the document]
chvck: I just hadn't got as far as doing that for first pass. Wanted to discuss if bits like that should be dumped as JSON or just text

memdx/ops_core.go
```diff
@@ -9,25 +9,32 @@ import (
 type OpsCore struct {
 }
 
-func (o OpsCore) decodeErrorContext(resp *Packet, err error) error {
+func (o OpsCore) decodeErrorContext(resp *Packet, err error, dispatchedTo string, dispatchedFrom string) error {
```
brett19, line 12: It might be useful to just pass the Dispatcher to this method, so we can capture whatever state we want in one place (DRY).  But this also works.

brett19 (approved):

memdx/ops_core.go
chvck: I'm going to merge this as is, just because I think that passing in an interface with a Dispatch() call would be a bit weird.

PR #6, opened 2023-02-15
Add support for durability to memdx

chvck:


memdx/durability.go
```diff
@@ -0,0 +1,20 @@
+package memdx
+
+// DurabilityLevel specifies the level to use for enhanced durability requirements.
+type DurabilityLevel uint8
+
+// TODO(chvck): Should this start
+const (
+	// DurabilityLevelNone indicates that no durability is needed.
+	DurabilityLevelNone = DurabilityLevel(0)
```
brett19, line 9: I think we previously ran into issues with not being able to detect the difference between 'Unspecified' and 'None'.  Perhaps we should make 0 be undefined instead (I think it's acceptable to need to map values for the purposes of gocb compatibility)?

memdx/ops_crud.go
```diff
@@ -33,6 +35,26 @@ func (o OpsCrud) encodeReqExtFrames(onBehalfOf string, buf []byte) (Magic, []byt
...
+	if durabilityLevel > 0 {
+		if durabilityLevelTimeout == 0 {
+			buf, err = AppendExtFrame(ExtFrameCodeReqDurability, []byte{byte(durabilityLevel)}, buf)
+			if err != nil {
+				return 0, nil, err
+			}
+		} else {
+			durabilityTimeoutMillis := durabilityLevelTimeout / time.Millisecond
+			if durabilityTimeoutMillis > 65535 {
+				durabilityTimeoutMillis = 65535
```
brett19, line 47: We should file a bug to make this an error instead in the future.

brett19 (approved):

memdx/ops_crud.go
chvck: GOCBC filed.

PR #7, opened 2023-02-17
Switch to using the authenticator pattern for authenticating.

brett19:


authenticator.go
```diff
@@ -0,0 +1,25 @@
+package core
+
+import "crypto/tls"
+
+type Authenticator interface {
+	GetClientCertificate(service ServiceType, hostPort string) (*tls.Certificate, error)
```
chvck, line 6: Am I right in saying that we never call this yet?

chvck (approved):

authenticator.go
brett19: Correct.

PR #8, opened 2023-02-20
Add support for compression

chvck:


chvck:
Intending on adding memdx.Dataflag to everything after this, to not make the review impossible

agent_test.go
```diff
@@ -84,6 +86,231 @@ func TestAgentBadCollection(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestAgentCrudCompress(t *testing.T) {
+	if !testutils.TestOpts.LongTest {
+		t.SkipNow()
+	}
+
+	agent := CreateDefaultAgent(t)
```
brett19, line 94: We should probably write a testutils helper that lets us see what's being dispatched from the crud component without actually needing a full agent (we essentially need to just mock the dispatcher I think?)

brett19 (approved):

PR #17, opened 2023-03-06
Update kvclient simplecall cancellation to return result if exists

chvck:


kvclient_ops.go
```diff
@@ -71,7 +71,9 @@ func kvClient_SimpleCall[Encoder any, ReqT any, RespT any](
 		return res.Result.(RespT), res.Err
 	case <-ctx.Done():
 		if !pendingOp.Cancel() {
-			<-resulter.Ch
+			res := <-resulter.Ch
```
chvck, line 74: Not certain about this but seems like we should return the actual result if we have it. Unsure about expectations as a user here though.

memdx/client.go
```diff
@@ -78,18 +78,26 @@ func (c *Client) registerHandler(handler DispatchCallback) uint32 {
...
 		c.lock.Unlock()
-		return false
+		return
 	}
 
-	delete(c.opaqueMap, opaqueID)
 	c.lock.Unlock()
-	return true
+
+	hasMorePackets := handler(nil, err)
```
brett19, line 92: I think we might need to make this enforce that hasMorePackets is false after an error is propagated rather than having it potentially keep the handler installed.

brett19 (changes requested):

PR #18, opened 2023-03-06
Add support for cancelling bootstrap to memdx

chvck:


memdx/pendingop_multi.go
```diff
@@ -0,0 +1,35 @@
...
+
+	for _, op := range ops {
+		op.Cancel(err)
+	}
+}
+
+func (op *multiPendingOp) Add(opToAdd PendingOp) {
+	op.lock.Lock()
+	err := op.cancelErr
+	if op.cancelErr != nil {
```
brett19, line 27: `if err != nil {` seems more sensible here?

memdx/pendingop_multi.go
```diff
@@ -0,0 +1,35 @@
...
+	ops       []PendingOp
+	lock      sync.Mutex
+	cancelErr error
+}
+
+func (op *multiPendingOp) Cancel(err error) {
+	op.lock.Lock()
+	op.cancelErr = err
+
+	ops := make([]PendingOp, len(op.ops))
```
brett19, line 15: Just a performance tidbit, but `cancelErr != nil` is technically a locking mechanism for the ops (you can't add ops after its non-nil), so you can probably elide the copy here, knowing that the ops list cannot change.

memdx/op_bootstrap.go
```diff
@@ -195,20 +232,33 @@ func (a OpBootstrap) Bootstrap(d Dispatcher, opts *BootstrapOptions, cb func(res
...
 	// maybeCallback must be invoked before any dispatches to ensure
 	// we still own all the state and to avoid racing those threads.
 	maybeCallback()
 
-	dispatchHello()
+	if err := dispatchHello(); err != nil {
+		return nil, err
+	}
+
+	return pendingOp, nil
```
brett19, line 258: I wonder if we should have a guard here (or even a comment) that talks about how the pendingOp would be broken if there were no ops added to it, but that the maybeCallback call _should_ prevent that.

memdx/op_bootstrap.go
```diff
@@ -79,18 +83,23 @@ func (a OpBootstrap) Bootstrap(d Dispatcher, opts *BootstrapOptions, cb func(res
 		}
 	}
 
-	dispatchHello = func() {
+	pendingOp := &multiPendingOp{}
```
brett19, line 86: I don't see anywhere that you remove ops from the multiPendingOp after they have completed.  I guess this is generally okay, since cancelling an already completed operation is a no-op.

brett19 (changes requested):

PR #33, opened 2023-03-08
Remove bucketname from agent reconfigure options

chvck:


brett19:
Closing this PR as we should be able to support Reconfiguring the bucket name, although only in the BucketName="" -> BucketName!="" case.

PR #38, opened 2023-03-08
Fix missing bootstrap

brett19:


agent.go
```diff
@@ -31,12 +31,14 @@ type Agent struct {
 	lock  sync.Mutex
 	state agentState
 
+	cfgWatcher  ConfigWatcher
```
chvck, line 34: Why do we have cfgWatcher, httpCfgWatcher, and memdCfgWatcher?

brett19: The concrete classes have different configuration and reconfiguration requirements, but the interface is useful for most cases.

brett19: Now that you mention it, I wonder if instead of exposing something like "WatchConfig()", if we should just expose the interfaces for all our internal components.

PR #40, opened 2023-03-09
Improved kv pool

brett19:


kvclient.go
```diff
@@ -202,33 +209,60 @@ func NewKvClient(ctx context.Context, opts *KvClientConfig) (*kvClient, error) {
...
 }
 
-func (c *kvClient) Reconfigure(ctx context.Context, opts *KvClientConfig) error {
-	if opts == nil {
-		return nil
+func (c *kvClient) Reconfigure(config *KvClientConfig, cb func(error)) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if config == nil {
```
chvck, line 216: This should probably be before the lock, but that's being picky.

kvclient.go
```diff
@@ -202,33 +209,60 @@ func NewKvClient(ctx context.Context, opts *KvClientConfig) (*kvClient, error) {
...
+		return errors.New("client config after reconfigure did not match new configuration")
 	}
 
-	if opts.SelectedBucket != "" {
-		if c.bucket != "" {
-			return placeholderError{"cannot perform select bucket on an already bucket bound kvclient"}
+	go func() {
+		if selectBucketName != "" {
+			err := c.SelectBucket(context.Background(), &memdx.SelectBucketRequest{
+				BucketName: config.SelectedBucket,
```
chvck, line 251: Should probably be selectBucketName, that or make selectBucketName a bool instead.

kvclient_test.go
```diff
@@ -32,19 +32,25 @@ func TestKvClientReconfigureBucket(t *testing.T) {
 	}
 
 	cli, err := NewKvClient(context.Background(), &KvClientConfig{
-		Logger:        logger,
```
chvck, line 35: Why? We should probably add not remove logging?

kvclient.go
brett19: Fixed.

kvclient.go
brett19: Fixed.

kvclient_test.go
brett19: We didn't actually remove it.  The Logger moved to KvClientOptions instead of KvClientConfig, and thus it's 'inherited' from the KvClientPool (as KvClientPool inherits from KvClientManager) rather than it being specified independently here.

PR #47, opened 2023-03-15
Update GetaMetaResult.Deleted to be a bool

chvck:


memdx/ops_crud.go
```diff
@@ -1300,7 +1300,7 @@ func (o OpsCrud) GetMeta(d Dispatcher, req *GetMetaRequest, cb func(*GetMetaResp
 			Value: resp.Value,
 			Cas:   resp.Cas,
 		}
-		res.Deleted = binary.BigEndian.Uint32(resp.Extras[0:])
+		res.Deleted = binary.BigEndian.Uint32(resp.Extras[0:]) == 1
```
brett19, line 1303: This should be != 0, since the server might want to add support for "deleted flags" or something.

brett19 (changes requested):

brett19 (approved):

PR #49, opened 2023-03-15
Implement Close for the kv client path

chvck:


brett19 (dismissed):

kvclientmanager.go
```diff
@@ -252,6 +253,27 @@ func (m *kvClientManager) GetClient(ctx context.Context, endpoint string) (KvCli
...
 }
+func (m *kvClientManager) Close() error {
+	state, err := m.getState()
+	if err != nil {
+		return err
+	}
+
+	m.logger.Info("closing kv client manager")
+
+	m.state.Store(nil)
```
brett19, line 264: There is a race here between m.getState() and m.state.Store(nil).  I think this would need to be a CompareAndSwap, or it should really be taking the same lock we use during reconfigure to avoid multiple concurrent calls from stepping on each others toes.

kvclientpool.go
```diff
@@ -49,9 +52,12 @@ type kvClientPool struct {
...
 	fastMap   AtomicPointer[kvClientPoolFastMap]
 
-	lock            sync.Mutex
-	config          KvClientPoolConfig
-	connectErr      error
+	lock       sync.Mutex
+	config     KvClientPoolConfig
+	connectErr error
+	closed     chan struct{}
+	// activeClients contains both current and defunct clients, this allows us to continue
```
brett19, line 59: Good call putting a comment on this.

kvclientpool.go
```diff
@@ -202,9 +229,11 @@ func (p *kvClientPool) startNewClientLocked() <-chan struct{} {
 		}
 
 		if err != nil {
-			p.logger.Warn("failed to create a new client connection", zap.Error(err))
-
-			p.connectErr = err
+			// Don't store the error if we've attempted to write an op after closing a connection,
+			// just recreate the client if required.
+			if !errors.Is(err, net.ErrClosed) {
```
brett19, line 234: This seems dubious.  I don't think we should be special-casing errors here.  Any error which leads the client to being unavailable should count an error from the perspective of being able to tell the user why their operations can't execute.

kvclientpool.go
```diff
@@ -439,7 +477,51 @@ func (p *kvClientPool) getClientSlow(ctx context.Context) (KvClient, error) {
...
+	p.fastMap.Store(nil)
+
+	// Close anyone waiting on this.
+	if p.needNoDefunctSigCh != nil {
+		close(p.needNoDefunctSigCh)
+		p.needNoDefunctSigCh = nil
+	}
+
+	// Just close all the clients in all the lists containing alive clients.
+	for _, client := range p.activeClients {
```
brett19, line 500: activeClients is more or less meant to be a 'cache' rather than something that is actively operated on.  It might make sense in that context to separately iterate the currentClients and defunctClients, and then update the 'activeClients cache' afterwards by clearing it.

brett19 (changes requested):

kvclientpool.go
```diff
@@ -431,6 +467,8 @@ func (p *kvClientPool) getClientSlow(ctx context.Context) (KvClient, error) {
 		} else {
 			return nil, ctxErr
 		}
+	case <-p.closed:
+		return nil, ErrPoolClosed
```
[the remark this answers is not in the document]
chvck: We could, but that doesn't exist :)

kvclientpool.go
chvck: This wasn't actually supposed to make it this far, I was using this check to debug something.

kvclientpool.go
```diff
@@ -394,6 +418,13 @@ func (p *kvClientPool) GetClient(ctx context.Context) (KvClient, error) {
 }
 
 func (p *kvClientPool) getClientSlow(ctx context.Context) (KvClient, error) {
+	// Check if we're already closed and if so return an error.
+	select {
+	case <-p.closed:
```
brett19, line 423: There are some weird details about dealing with channels, and there is a performance impact to doing a select like this.  It would likely be better to have a `closed bool` protected by the lock, and then the closedSig is only used to watch the channel if we are going to be blocking for another signal.

kvclientmanager.go
```diff
@@ -252,6 +253,30 @@ func (m *kvClientManager) GetClient(ctx context.Context, endpoint string) (KvCli
 
 	return connProvider.GetClient(ctx)
 }
+func (m *kvClientManager) Close() error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	state, err := m.getState()
```
brett19, line 260: Since we are manually mutating the state in this method, it probably makes sense to directly use m.state.Load(), and do our own checks for nil.  This would be especially helpful for code-clarity in being able to understand that there is logic that's catching the case where the manager is already closed.

brett19 (changes requested):

memdx/errors.go
```diff
@@ -47,6 +47,8 @@ var (
 	ErrSubDocXattrUnknownVattrMacro        = errors.New("subdoc xattr unknown vattr macro")
 	ErrSubDocCanOnlyReviveDeletedDocuments = errors.New("subdoc can only revive deleted documents")
 	ErrSubDocDeletedDocumentCantHaveValue  = errors.New("subdoc deleted document cant have value")
+
+	ErrClosedInFlight = errors.New("connection closed whilst operation in flight")
```
chvck, line 51: Unsure this really needs to be a sentinel error.

kvclient.go
```diff
@@ -294,7 +313,26 @@ func (c *kvClient) HasFeature(feat memdx.HelloFeature) bool {
 
 func (c *kvClient) Close() error {
 	c.logger.Info("closing")
-	return c.cli.Close()
+	if !atomic.CompareAndSwapUint32(&c.closed, 0, 1) {
```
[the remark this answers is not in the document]
brett19: I would generally expect that you might use signalling to solve this.  Where Close() does "block" waiting for Run() to return through a signal, with the expectation that the goroutine would end immediately after signalling that it has completed.  Then here the `closed` var would signal whether you need to do the Close() call, but all calls to Close() would wait on the `closedSig chan struct{}` (such that multiple calls to Close() all wait for it to actually be done closing).

kvclient_ops.go
```diff
@@ -52,11 +53,14 @@ func kvClient_SimpleCall[Encoder any, ReqT any, RespT any](
 	req ReqT,
 ) (RespT, error) {
 	resulter := allocSyncCrudResulter()
+	var calledBack uint32
 
 	pendingOp, err := execFn(o, c.cli, req, func(resp RespT, err error) {
-		resulter.Ch <- syncCrudResult{
-			Result: resp,
-			Err:    err,
+		if atomic.CompareAndSwapUint32(&calledBack, 0, 1) {
```
brett19, line 59: We should probably have an `} else { panic() }` since this is a serious logic error on our part if it occurs.  We may also want to have this check only occur when we are running with --race or something, since assuming gocbcorex is written correctly, this should never occur.

chvck: I think that logger.DPanic() would probably the answer there.

brett19 (dismissed):

brett19 (approved):
Mostly looks good, although I think there are new bugs to file related to this still based on a discussion I had with @chvck

PR #66, opened 2023-03-23
Close response bodies for mgmt collections operations

chvck:


testutils/utils.go
```diff
@@ -120,6 +121,9 @@ func SetupTests(m *testing.M) {
 
 	result := m.Run()
 
+	// We need to close the transport used by the default client once tests complete, otherwise the transport
+	// will leak go routines.
+	http.DefaultClient.CloseIdleConnections()
```
brett19, line 126: This is interesting since it implies that we must be _using_ the DefaultClient, which I don't think we should be?

chvck: It's in tests only. We do use the default transport in tests.

brett19: Ahh yes.  +1

brett19 (approved):

brett19 (approved):

PR #70, opened 2023-03-28
Close config watcher on agent close

chvck:


agent.go
```diff
@@ -38,8 +38,9 @@ type Agent struct {
 	retries     RetryManager
 	vbRouter    VbucketRouter
 
-	httpCfgWatcher *ConfigWatcherHttp
-	memdCfgWatcher *ConfigWatcherMemd
+	httpCfgWatcher   *ConfigWatcherHttp
+	memdCfgWatcher   *ConfigWatcherMemd
+	cfgWatcherCancel func()
```
brett19, line 43: I wonder if we should just have an 'agentBackgroundContext' that covers all background work that an Agent is directly performing.  I guess that doesn't work well for waiting for them to all finish though.

brett19 (approved):

PR #71, opened 2023-03-28
Store http transport on agent so that it can be closed

chvck:


agent.go
```diff
@@ -282,7 +287,9 @@ func (agent *Agent) genAgentComponentConfigsLocked() *agentComponentConfigs {
 	var mgmtEndpoints []string
 	var queryEndpoints []string
 	var searchEndpoints []string
-	if agent.state.tlsConfig == nil {
+
+	tlsConfig := agent.state.tlsConfig.Get()
```
brett19, line 291: We already have a lock on the tlsConfig, so we shouldn't need to make it an atomic.  I can see the atomicTlsConfig being useful if we want to be able to reuse the same transport through multiple reconfigure's though.  By having the transport hold a reference to an atomicTlsConfig, and then reconfigure just updates the atomic.  That being said, I think the better approach here long-term is that we keep track of the transport as this PR does, and then _if_ the TLS config has changed, we replace the transport, but otherwise reuse the same transport.  Otherwise we end up unnecessarily preventing connection reuse because of new configurations that had no impact on the transport.

brett19 (approved):

PR #72, opened 2023-03-28
Fail tests if any goroutine leaks are detected

chvck:


testutils/utils.go
```diff
@@ -138,6 +138,7 @@ func SetupTests(m *testing.M) {
 	if finalGoroutineCount != initialGoroutineCount {
```
brett19, line 138: Maybe the above loop should be using assert.Eventually?

testutils/utils.go
```diff
@@ -138,6 +138,7 @@ func SetupTests(m *testing.M) {
...
 		log.Printf("Detected a goroutine leak (%d before != %d after)", initialGoroutineCount, finalGoroutineCount)
 		pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
+		result = 1
```
brett19, line 141: We should probably not be overwriting an already non-zero result here?  Unless test.Run() only ever returns 0/1.

brett19 (approved):

testutils/utils.go
chvck: I believe that tests only return 0/1. 2 Can be returned if testing is interrupted.

PR #83, opened 2023-04-11
Refactored HTTP OBO handling to actually work.

brett19:


cbhttpx/base.go
```diff
@@ -32,13 +41,23 @@ func (h RequestBuilder) NewRequest(
...
-	if onBehalfOf != "" {
-		req.Header.Set("cb-on-behalf-of", onBehalfOf)
-	}
-
 	if h.BasicAuthUser != "" || h.BasicAuthPass != "" {
 		req.SetBasicAuth(h.BasicAuthUser, h.BasicAuthPass)
 	}
 
+	if onBehalfOf != nil {
+		if onBehalfOf.Password != "" {
```
chvck, line 49: I think it'd probably be nice if this was something like onBehalfOf.HasPassword() or onBehalfOf.UseBasicAuth(). Just be a bit easier to read.

chvck (dismissed):

cbhttpx/base.go
brett19: I mostly did it this way because this isn't really the 'final' logic here.  Realistically it should be switching whether to use the password or domain based on server support rather than what was provided by the user.

PR #92, opened 2023-04-17
Remove double callback check for bootstrap

chvck:


memdx/client.go
```diff
@@ -129,13 +126,17 @@ func (c *Client) dispatchCallback(pak *Packet) error {
 		return nil
 	}
 
+	// We delete and then readd the opaque if required.
+	// We do this to prevent a race with Close where both sides can end up
+	// seeing the opaque in the opaque map at the same time.
+	delete(c.opaqueMap, pak.Opaque)
```
brett19, line 132: As discussed on slack, this has issues with racing between handlers and Cancel() calls.  Primarily that if Cancel() is invoked _while_ the handler is running, it won't be in the opaque map and thus does not get cancelled appropriately.

brett19 (changes requested):

iamrajiv (changes requested):
@chvck, if the double callback check for bootstrap is removed, how will it handle a callback?

chvck:
> @chvck, if the double callback check for bootstrap is removed, how will it handle a callback?

Bootstrap (and any other op for that matter) should only be calling the callback provided to the operation a single time. Calling the callback more once is a bug. This change makes it so that there is no protection against calling the callback more than once, and will trigger a panic (or error log in production) at a higher level where there is a check for this occurring if it does.

iamrajiv:
@chvck, I understand that you removed the check `atomic.CompareAndSwapUint32(&calledBack, 0, 1)` to prevent the callback from being called multiple times, and instead added a panic to handle that scenario. However, I'm curious about the bug that was introduced by using `atomic.CompareAndSwapUint32(&calledBack, 0, 1)`. Can you provide more information on that?

chvck:
> However, I'm curious about the bug that was introduced by using atomic.CompareAndSwapUint32(&calledBack, 0, 1). Can you provide more information on that?

There was no bug introduced as such. It could just be masking problematic behaviour so needs to be removed.

iamrajiv:
@chvck, since we're using `atomic.CompareAndSwapUint32(&calledBack, 0, 1)` in multiple places in the codebase, do you think we should refactor those sections similarly to what we did here?

chvck:
> @chvck, since we're using `atomic.CompareAndSwapUint32(&calledBack, 0, 1)` in multiple places in the codebase, do you think we should refactor those sections similarly to what we did here?

No, I believe that we only use it in one place after this change. The usage there is ok.

iamrajiv:
Hey @chvck, what do you think about this code https://github.com/couchbase/gocbcorex/blob/542841967806bb73a6284f3a4b2d0328fab54b14/kvclient.go#L299?

iamrajiv:
@chvck, what do you think about the following code?

https://github.com/couchbase/gocbcorex/blob/542841967806bb73a6284f3a4b2d0328fab54b14/kvclient.go#L299?

iamrajiv:
@chvck, what do you think about the following code?

https://github.com/couchbase/gocbcorex/blob/542841967806bb73a6284f3a4b2d0328fab54b14/kvclient.go#L299

chvck:
> @chvck, what do you think about the following code?
> 
> https://github.com/couchbase/gocbcorex/blob/542841967806bb73a6284f3a4b2d0328fab54b14/kvclient.go#L299

I think that it's correct and necessary. That `Close` function can be called from multiple goroutines concurrently and that check prevents the client from trying to close the connection twice.

brett19 (changes requested):
Marking this as Request Changes so that we don't accidentally merge it until we've fixed the underlying problem that required the workaround we are removing in this commit.

brett19:
This should have been resolved by another commit a while back.

PR #97, opened 2023-04-20
Add orphaned response and connection close handlers to kv stack

chvck:


kvclient.go
```diff
@@ -315,3 +319,21 @@ func (c *kvClient) RemoteAddress() string {
...
+		"orphaned response encountered",
+		zap.String("opaque", strconv.Itoa(int(packet.Opaque))),
+		zap.String("opcode", packet.OpCode.String()),
+	)
+}
+
+func (c *kvClient) handleConnectionClose(err error) {
+	// Just mark ourselves as closed. The connection is already
+	// closed so there's no actual work to do, and we might already actually be closed.
+	atomic.StoreUint32(&c.closed, 1)
```
chvck, line 334: We could CAS this and only call upstream if we aren't already closed. Thoughts?

PR #99, opened 2023-04-22
Added status code to memdx server error structure.

brett19:


chvck (approved):

chvck:
Looks like a legit test fail.

PR #102, opened 2023-04-25
Return error when vbucketMap is empty and handle panic

boseabhishek:


chvck:
I think that this is a good change but we should investigate what's causing this panic first, we risk masking a bug here.

boseabhishek:
> I think that this is a good change but we should investigate what's causing this panic first, we risk masking a bug here.

Yes, makes sense.

vbucketmap.go
```diff
@@ -29,10 +29,15 @@ func (vbMap VbucketMap) NumReplicas() int {
 	return vbMap.numReplicas
 }
 
-func (vbMap VbucketMap) VbucketByKey(key []byte) uint16 {
+func (vbMap VbucketMap) VbucketByKey(key []byte) (uint16, error) {
```
brett19, line 32: This change modifies the vbucketmap code to handle the case where there is no vbucket map, but wouldn't that imply this entire structure shouldn't exist, rather than it existing but being invalid?

brett19 (changes requested):

vbucketmap.go
boseabhishek: If I understand your comment correctly, it's logical that this issue will be probably caught higher up the chain. But, imo this creates a defence line and avoids any probable panics. Inside func `NodeByKey` [here](https://github.com/couchbase/gocbcorex/pull/102/files/72840d5a10b0d30e154364c15a77d5c1357e4ffa#diff-5b0009b214f09d2ff2d6f7c7b1c46d3a45fe23d9a14de9872f2bd41bcb75d127R109), I have tried checking that before it get's there.

But, I guess your point is when/before `VbucketRoutingInfo` is getting populated?

brett19:
Hey @boseabhishek , are there any updates on this?

brett19:
I'm going to close this in favour of https://github.com/couchbase/gocbcorex/pull/121 for now.  Feel free to re-open this if you think the alternative PR doesn't solve the underlying issue.

boseabhishek:
Sorry I'm swamped with other stuff @brett19 and completely forgot about it. Sure, it makes sense.

PR #104, opened 2023-04-25
Bucketwarmup

chvck:


configbootstrap_http.go
```diff
@@ -103,37 +103,63 @@ func configBootstrapHttp_bootstrapOne(
...
-		return configBootstrapHttp_bootstrapOne(
-			ctx,
-			w.httpRoundTripper,
-			w.endpoints[0],
-			w.userAgent,
-			w.authenticator,
-			w.bucketName,
-		)
-	}
+	for {
```
brett19, line 106: We should probably clean this up so that it's just 1 case (rather than having an `if len() == 1` block).  I also think that perhaps we should consider having configBootstrapHttp_bootstrapOne consider an empty vbucket map as something that it retries internally rather than having it return a parsedConfig that's invalid?

brett19 (changes requested):

brett19 (approved):
Approved, although we might want a comment around the if block for BucketName==""||vbucketMap.IsValid() that clarifies that this only happens transiently when we are warming up, and thus we just retry.  We may also want to implement a short sleep before trying again.

PR #105, opened 2023-04-25
Retry vbucket map outdated errors in default retry manager

chvck:


retrymanager_default.go
```diff
@@ -30,7 +30,7 @@ type retryControllerDefault struct {
 
 func (rc *retryControllerDefault) isRetriableError(err error) bool {
 	// Implement the default classification of retriable errors...
-	return errors.Is(err, memdx.ErrTmpFail)
+	return errors.Is(err, memdx.ErrTmpFail) || errors.Is(err, ErrVbucketMapOutdated)
```
brett19, line 33: I believe we should also be retrying on ErrCollectionMapOutdated.

brett19 (changes requested):

brett19 (approved):

brett19:
Approved, though we should file a bug to include ErrCollectionMapOutdated in the retriable errors if it's not added to this PR before merge.

PR #108, opened 2023-04-26
Update memdx tests missing vbucket id

chvck:


brett19 (changes requested):
If hard-coding this to 0 isn't sufficient, perhaps we should be generating the vbucket-id based on the topology to avoid NVM errors?

brett19:
Based on a conversation with @chvck , I understand why vbucket 1 is picked here, but we should probably adjust the tests to use `defaultTestVbucketID` instead of hardcoding that 1 value in multiple places.

brett19 (approved):

brett19:
Rebased to master and adjusted `VbucketID: 1` to `VbucketID: defaultTestVbucketID`.

PR #117, opened 2023-05-09
Always send explain as a part of search query requests

chvck:


cbsearchx/query_options.go
```diff
@@ -120,9 +120,7 @@ func (o *QueryOptions) encodeToJson() (json.RawMessage, error) {
 		}
 		m["ctl"] = encoder.EncodeField(control)
 	}
-	if o.Explain {
-		m["explain"] = encoder.EncodeField(o.Explain)
-	}
+	m["explain"] = encoder.EncodeField(o.Explain)
```
brett19, line 123: Is this the right approach, or is it `!o.Explain { Encode(!xxx) }`, since the server default is false?

brett19 (approved):

cbsearchx/query_options.go
chvck: I suppose that would save a small amount of bytes in the payload but otherwise no much benefit? I think this approach also insulates us to if the server decided to change the behaviour.

brett19:
+1

cbsearchx/query_options.go
brett19: +1

PR #121, opened 2023-05-11
ING-485: Added handling for invalid vbucket maps.

brett19:


boseabhishek (commented):
LGTM.. just one little thing, though it might never get there - imo we should still guard the last line of defence here just to avoid divide by zero panic [here](https://github.com/couchbase/gocbcorex/blob/3fc96e2ca8e34091882d1f6102e7ef07e880e777/vbucketmap.go#L35)

brett19:
@boseabhishek: We do guard against a divide by zero there, it just happens earlier now, preventing the creation of the VbucketMap object at all:
https://github.com/couchbase/gocbcorex/pull/121/files#diff-5b0009b214f09d2ff2d6f7c7b1c46d3a45fe23d9a14de9872f2bd41bcb75d127R14

boseabhishek (dismissed):

chvck (approved):

PR #124, opened 2023-05-15
Added mechanism to turn on packet logging using env var.

brett19:


memdx/client.go
```diff
@@ -2,11 +2,14 @@ package memdx
 
 import (
 	"errors"
+	"os"
 	"sync"
 
 	"go.uber.org/zap"
 )
 
+var enablePacketLogging bool = os.Getenv("GCBCX_PACKET_LOGGING") != ""
```
chvck, line 11: Does this mean that it can't be dynamically toggled?

brett19: At the moment, you could only turn it on/off globally at application startup.

chvck (approved):

PR #125, opened 2023-05-18
When cached statment prepared query fails then reset prepared

chvck:


brett19 (approved):
Approved, but also filed an issue to make this handling more robust:
https://github.com/couchbase/gocbcorex/issues/126

PR #133, opened 2023-05-20
#132: Added helpers for subdocument projection.

brett19:


helpers/subdocpath/subdocpath.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"strconv"
+)
+
+type Element struct {
+	Key   string
+	Index int
+}
+
+func Parse(path string) ([]Element, error) {
```
dnault, line 14: Does this method handle escape sequences? For example:

    {
      "x.y": "a",  // subdoc path "`x.y`" or  "x`.`y"
      "x`y": "b"   // subdoc path "x``y" or "`x``y`"
    }

I think the escaping rule is that a backtick starts an escaped region, unless it's immediately followed by another backtick, in which case the backtick pair represents a literal backtick.

dnault (changes requested):

helpers/subdocpath/subdocpath.go
brett19: Good point.  I'll go look this up in the sub-doc spec!

brett19: I've updated the PR to correctly handle backtick escaped paths.

dnault (approved):

PR #143, opened 2023-06-08
Pin range scan continue/cancel to same connection as create

chvck:


brett19 (changes requested):
From Slack discussion:
```
type RangeScan interface {
    Continue(ctx context.Context, opts *RangeScanContinueOptions, dataCb func(RangeScanContinueDataResult)) (*RangeScanContinueResult, error)
    Cancel(...) (...)
}

func (cc *CrudComponent) RangeScan(ctx context.Context, opts *memdx.RangeScanCreateOptions) (RangeScan, *memdx.RangeScanCreateResult, error)
```

brett19 (approved):
I've pushed a small change that removed the ID parts of KvClient, since they are not used anymore now that we directly capture the KvClient itself.

PR #150, opened 2023-07-11
Initial port of transactions from gocbcore

brett19:


transactionsx/error_transactions.go
```diff
@@ -291,9 +291,7 @@ func marshalErrorToJSON(err error) json.RawMessage {
 		}
 	}
 
-	data, err := json.Marshal(err.Error())
-	if err != nil {
-		logWarnf("Failed to marshal error: %v", err)
-	}
+	// Marshalling a string cannot fail
```
chvck, line 294: I can see this failing linting at some future point

transactionsx/transaction.go
```diff
@@ -241,32 +226,29 @@ type TransactionMutableItemMeta struct {
...
-	Cas   Cas
+	Cas   uint64
 }
 
-// TransactionGetCallback describes a callback for a completed Get or GetOptional operation.
-type TransactionGetCallback func(*TransactionGetResult, error)
-
 // Get will attempt to fetch a document, and fail the transaction if it does not exist.
-func (t *Transaction) Get(opts TransactionGetOptions, cb TransactionGetCallback) error {
+func (t *Transaction) Get(ctx context.Context, opts TransactionGetOptions) (*TransactionGetResult, error) {
```
chvck, line 241: Should ctx belong to the transaction rather than individual ops? Although pretty sure it's an anti-pattern to store context so perhaps not...

transactionsx/transaction.go
```diff
@@ -471,9 +447,3 @@ func (t *Transaction) UpdateState(opts TransactionUpdateStateOptions) {
 
 	t.attempt.UpdateState(opts)
 }
-
-// Logger returns the logger used by this transaction.
-// Uncommitted: This API may change in the future.
-func (t *Transaction) Logger() TransactionLogger {
```
chvck: Is there a reason to not expose this anymore?

transactionsx/transactionattempt.go
```diff
@@ -116,7 +134,8 @@ func (t *transactionAttempt) FinalErrorToRaise() TransactionErrorReason {
 }
 
 func (t *transactionAttempt) UpdateState(opts TransactionUpdateStateOptions) {
-	t.logger.logInfof(t.id, "Updating state to %s", opts)
+	t.logger.Info("Updating state", zap.Stringer("opts", opts))
```
chvck, line 137: Nothing sensitive in these options?

transactionsx/transactions.go
```diff
@@ -1,91 +1,118 @@
...
-	config      TransactionsConfig
-	cleaner     TransactionsCleaner
-	lostCleanup lostTransactionCleaner
+	config TransactionsConfig
 }
 
 // InitTransactions will initialize the transactions library and return a TransactionsManager
 // object which can be used to perform transactions.
 func InitTransactions(config *TransactionsConfig) (*TransactionsManager, error) {
-	logInfof("Initializing transactions: %s", config)
```
chvck: I feel like this log line is actually useful

chvck: I see below

chvck (approved):
Nothing jumps out at me as being problematic here

PR #156, opened 2023-09-13
Remove check blocking ephemeral buckets from being created

chvck:


cbmgmtx/mgmt.go
```diff
@@ -519,8 +519,6 @@ func (h Management) encodeMutableBucketSettings(posts *url.Values, opts *Mutable
 		} else {
 			posts.Add("replicaIndex", "1")
 		}
-	} else {
```
brett19, line 522: I think this should still be here, it's just that it's missing the check for `ReplicaIndexDisable`?
```
	if opts.BucketType != BucketTypeEphemeral {
		if opts.ReplicaIndexDisabled {
			posts.Add("replicaIndex", "0")
		} else {
			posts.Add("replicaIndex", "1")
		}
	} else {
		if opts.ReplicaIndexDisabled {
			return errors.New("cannot specify ReplicaIndexDisabled for Ephemeral buckets")
		}
	}
```

DemetrisChr (approved):

PR #158, opened 2023-10-03
Agentmanager

chvck:


chvck:
I'm not entirely sure about this internal service agent manager implementation but it is resilient to rebalance and does track buckets on the cluster. It'd be nicer if there was a single endpoint that we could poll for management endpoints and buckets but there isn't.

internalservice_agentmanager.go
```diff
@@ -0,0 +1,331 @@
...
+
+	topologyCfgWatcher *ConfigWatcherHttp
+}
+
+type bucketDescriptor struct {
+	Name string
+	UUID string
+}
+
+func CreateInternalServiceAgentManager(ctx context.Context, opts InternalServiceAgentManagerOptions) (*InternalServiceAgentManager, error) {
```
brett19, line 65: We should probably name it based on it's behavior rather than our presumptions about its use.  InternalServiceAgentManager implies its only useful for internal services, naming it something like "AllBucketsAgentManager" or... "WatchingAgentManager" is probably more descriptive (I dislike those names though, there _must_ be something better).

internalservice_agentmanager.go
```diff
@@ -0,0 +1,331 @@
...
+		// Cancel the context and wait for the configCh to be closed.
+		cancel()
+		<-stoppedBucketSig
+		<-stoppedTopoSig
+	}
+
+	bucketCh := m.bucketsWatcher.Watch(ctx, streamWatcherHttp_streamBuckets)
+
+	go func() {
+		for bucket := range bucketCh {
```
brett19, line 164: `bucket` here is actually a list of buckets I think?  This should probably be called `buckets`.

internalservice_agentmanager.go
```diff
@@ -0,0 +1,331 @@
...
+	m.bucketsWatcher.Reconfigure(&StreamWatcherHttpConfig{
+		HttpRoundTripper: m.state.httpTransport,
+		Endpoints:        mgmtEndpoints,
+		UserAgent:        m.userAgent,
+		Authenticator:    m.state.authenticator,
+	})
+}
+
+func (m *InternalServiceAgentManager) handleBuckets(buckets []bucketDescriptor) {
+	m.bucketsLock.Lock()
```
brett19, line 222: We should probably file a bug to cover the fact that this implementation still has the same problem the original did where it ends up holding a 'global' lock when creating agents, making it so that you cant fetch the bucketA while bucketB is being connected.  This probably is trivially solvable by using a similar approach as we have with collections, with the slow map, and a fast lookup table.

brett19 (dismissed):
Few comments, but approving as this implementation is already significantly better than the existing one.

internalservice_agentmanager.go
chvck: https://github.com/couchbase/gocbcorex/issues/159

brett19 (approved):

PR #161, opened 2023-10-05
Ing588

chvck:


bucketswatcher_http.go
```diff
@@ -0,0 +1,300 @@
...
+	agent, ok = bw.agents[bucketName]
+	bw.lock.Unlock()
+	if !ok {
+		return nil, cbmgmtx.ErrBucketNotFound
+	}
+
+	return agent, nil
+}
+
+func (bw *BucketsWatcherHttp) Watch() {
```
chvck, line 146: This is a bit of a departure of our usual watcher API, which I don't love but kind of needs to be how it is.

bucketswatcher_http.go
```diff
@@ -0,0 +1,300 @@
...
+	agent, ok := bw.agents[bucketName]
+	if ok {
+		bw.lock.Unlock()
+		return agent, nil
+	}
+	bw.lock.Unlock()
+
+	bw.logger.Debug("Bucket unknown, attempting manual update", zap.String("name", bucketName))
+
+	buckets, err := bw.manuallyFetchBuckets(ctx)
```
chvck, line 128: Risk here that concurrent calls to GetAgent for an unknown bucket will fire off a bunch of requests, but if we want to ensure we're always up to date AFTER the user makes the call then it has to be this way.

chvck:
This still has a large lock but is a step toward removing that.

bucketswatcher_http.go
brett19: How do you see it as a departure?  Why does it need to be different?

bucketswatcher_http.go
brett19: The best way to solve this I think is that we should record when the user request came in (or at least, record when they invoked `GetAgent`).  Then we deduplicate the manual fetches from ns_server, so that only a single request can be outstanding at once.  When the outstanding request resolves, we resolve any `GetAgent` calls which were before the request was sent.  If there are still pending `GetAgent` calls which need newer information, we can re-dispatch the manual fetch with a newer timestamp which should 'clear the queue' so to speak.

bucketswatcher_http.go
chvck: Our watchers typically take a context which is used for closing the watcher. For this watcher that doesn't work though as the buckets watcher owns a set of agents which need to be closed and doing that should block the caller. If we were to close the agents on context cancel we'd have no way of blocking the caller.

brett19 (approved):

PR #162, opened 2023-10-05
ING-591: Do not check if ReplicaIndexDisabled for ephemeral buckets

chvck:
Motivation
----------
We currently check if ReplicaIndexDisabled is set, and for ephemeral buckets return an error if so. We should just ignore this field for ephemeral buckets. It's completely reasonable for a user to set the value to true as the server will not allow the replicaIndex field to be set, so it would be expected that setting the field to true could lead to the parameter being omitted in the request. As we cannot distinguish false and not set for the field we cannot take any taken there either.

### DIFF
--- a/cbmgmtx/mgmt.go
+++ b/cbmgmtx/mgmt.go
@@ -570,7 +570,7 @@ func (h Management) decodeMutableBucketSettings(data *bucketSettingsJson) (*Muta
 type BucketSettings struct {
 	MutableBucketSettings
 	ConflictResolutionType ConflictResolutionType
-	ReplicaIndexDisabled   bool
+	ReplicaIndex           bool
 	BucketType             BucketType
 	StorageBackend         StorageBackend
 }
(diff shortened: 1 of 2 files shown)


brett19:
Should we instead be using an optional field here?  Or inverting the logic of this field so that we adjust the 'replicaIndexDisabled' based on the 'replicaIndexes' field being set to 0 (or unset for ephemeral buckets, which would default to none instead of whatever the default is for non-ephemeral buckets)?

chvck:
> Should we instead be using an optional field here? Or inverting the logic of this field so that we adjust the 'replicaIndexDisabled' based on the 'replicaIndexes' field being set to 0 (or unset for ephemeral buckets, which would default to none instead of whatever the default is for non-ephemeral buckets)?

I've made it optional. I dislike the idea of setting replicaIndex based on the number of replicaIndexes - it's entirely viable for a user to want to disable replica indexes being enabled without actually modifying the number of replicas.

cbmgmtx/mgmt.go
```diff
@@ -583,7 +579,8 @@ func (h Management) decodeMutableBucketSettings(data *bucketSettingsJson) (*Muta
 	settings := MutableBucketSettings{}
 
 	settings.FlushEnabled = data.Controllers.Flush != ""
-	settings.ReplicaIndexDisabled = !data.ReplicaIndex
+	replicaIndexDisabled := !data.ReplicaIndex
+	settings.ReplicaIndexDisabled = &replicaIndexDisabled
```
programmatix, line 583: So this is the JSON from the server?  Can it not contain a ReplicaIndex field, and what happens in that case?  I'd expect the final Protostellar result to not have that field set.  But hmm - just saw that ListBucketsResponse has most fields non-optional.  Is that what we want?

chvck: replicaIndex is not present for ephemeral buckets. The only other field I can see not always being present is flush. But flush is either set to a uri or not set, so we'd have it as always true or nil - which is a bit odd IMO.

I think it could be argued that ephemeral not having it set is the same as false too. I think this change is actually wrong and we should just ignore the field on the create side for ephemeral buckets.

chvck:
I've inverted the field so that it's ReplicaIndex, and we now return an error if the bucket is ephemeral and ReplicaIndex is true. If the field is false and bucket is ephemeral then we just don't send the parameter in the form. I think that this is probably the most idiomatic approach.

Westwooo (approved):
lgtm

PR #164, opened 2023-10-09
ING-595: Add handle for doc too big status code

chvck:


memdx/errors.go
```diff
@@ -15,6 +15,7 @@ var (
 	ErrCollectionsNotEnabled               = errors.New("collections not enabled")
 	ErrDocNotFound                         = errors.New("document not found")
 	ErrDocExists                           = errors.New("document already exists")
+	ErrDocTooBig                           = errors.New("document too big")
```
DemetrisChr, line 18: The RFC has this with the name `ValueTooLarge`, is the deviation from the RFC here intentional?

chvck: Good spot, I forgot that the RFC exists.

DemetrisChr (approved):

PR #165, opened 2023-10-10
ING-572: Add a fast map to buckets tracking http

chvck:
Motivation
----------
At the moment if a user calls GetAgent when the tracker is handling detection of new buckets then the call will always block until all new agents are created and old ones destroyed. We should add a fast map so that requests for agents that we already know about can be resolved without waiting for a global lock.

### DIFF
--- a/bucketswatcher_http.go
+++ b/bucketswatcher_http.go
@@ -47,13 +47,15 @@ type BucketsWatcherHttp struct {
 	userAgent string
 	closed    bool
 
-	agents    map[string]*Agent
-	lock      sync.Mutex
+	fastMap   AtomicPointer[map[string]*Agent]
+	slowMap   map[string]*Agent
+	slowLock  sync.Mutex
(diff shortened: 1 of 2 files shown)


bucketswatcher_http.go
```diff
@@ -137,31 +139,35 @@ func (bw *BucketsWatcherHttp) manualUpdateThread() {
 
 		bw.handleBuckets(ctx, buckets)
 
-		bw.lock.Lock()
+		bw.needsBucketLock.Lock()
```
Westwooo, line 142: Why can't we use `statelock` here?

Westwooo (approved):
lgtm, just one question

bucketswatcher_http.go
chvck: Just a minor optimization so that updating state doesn't block access to the needsBucketChan as they're not related.

PR #166, opened 2023-10-10
ING-597: Add create agent timeout for buckets tracking manager

chvck:
Motivation
----------
The buckets tracking manager does not deadline new agent creation, this could mean that GetAgent could block indefinitely or our watcher thread could also become blocked - which in turn could also block GetAgent.

Changes
-------
Added a CreateAgentTimeout to buckets tracking manager options. Use the new timeout when creating a context to pass into CreateAgent.

### DIFF
--- a/bucketstracking_agentmanager.go
+++ b/bucketstracking_agentmanager.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"net/http"
 	"sync"
+	"time"
 
 	"go.uber.org/zap"
 )
@@ -24,6 +25,7 @@ type BucketsTrackingAgentManagerOptions struct {
 	CompressionConfig  CompressionConfig
 	ConfigPollerConfig ConfigPollerConfig
 	HTTPConfig         HTTPConfig
+	CreateAgentTimeout time.Duration
 }
 
 type bucketsTrackingAgentManagerState struct {
@@ -43,6 +45,7 @@ type BucketsTrackingAgentManager struct {
 	compressionConfig  CompressionConfig
 	configPollerConfig ConfigPollerConfig
 	httpConfig         HTTPConfig
+	createAgentTimeout time.Duration
 
 	state *bucketsTrackingAgentManagerState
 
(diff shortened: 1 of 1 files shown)


chvck:
We should probably think about deadlining the manual update http request too.

bucketstracking_agentmanager.go
```diff
@@ -91,12 +91,21 @@ func CreateBucketsTrackingAgentManager(ctx context.Context, opts BucketsTracking
 		zap.Any("bootstrapConfig", bootstrapConfig),
 		zap.String("networkType", networkType))
 
+	createAgentTimeout := opts.CreateAgentTimeout
+	if createAgentTimeout == 0 {
+		createAgentTimeout = 7 * time.Second
```
Westwooo, line 96: Why 7 seconds?

chvck: 7 seconds is what we typically use as the connection timeout in the SDK, I'm unsure where that value originated.

streamwatcher_http.go
```diff
@@ -139,13 +145,12 @@ func (w *StreamWatcherHttp[StreamT]) watch(
 	close(outCh)
 }
 
-func (w *StreamWatcherHttp[StreamT]) Reconfigure(Buckets *StreamWatcherHttpConfig) error {
+func (w *StreamWatcherHttp[StreamT]) Reconfigure(cfg *StreamWatcherHttpReconfigureConfig) error {
 	w.lock.Lock()
 	w.state = &streamWatcherHttpState{
-		httpRoundTripper: Buckets.HttpRoundTripper,
-		endpoints:        Buckets.Endpoints,
-		userAgent:        Buckets.UserAgent,
```
Westwooo, line 147: Why are we removing UserAgent from here?

Westwooo (commented):
lgtm, got a couple of questions

bucketswatcher_http.go
```diff
@@ -0,0 +1,350 @@
...
+
+type BucketsWatcherHttp struct {
+	stateLock sync.Mutex
+	state     *bucketsWatcherHttpState
+	logger    *zap.Logger
+	userAgent string
+	closed    bool
+
+	fastMap   AtomicPointer[map[string]*Agent]
+	slowMap   map[string]*Agent
```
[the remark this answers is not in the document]
Westwooo: Can a fast map be used without a lock, whereas a slow map cannot?

streamwatcher_http.go
chvck: useragent shouldn't be a part of reconfiguration.

Westwooo (approved):

PR #167, opened 2023-10-10
ING-590: Return invalid args error on write packet validation fail

chvck:
Motivation
----------
On the write side of writing to the connection anything which causes the validation of the packet to fail should be an invalid args error rather than a protocol error. It means that there is an issue in the user's packet rather than a protocol issue like there would be for a response packet.

Changes
-------
Updated WritePacket to return invalid argument errors on packet validation failure.

### DIFF
--- a/memdx/packetwriter.go
+++ b/memdx/packetwriter.go
@@ -26,31 +26,31 @@ func (pw *PacketWriter) WritePacket(w io.Writer, pak *Packet) error {
 
 	if pak.Magic == MagicReq || pak.Magic == MagicRes {
 		if extFramesLen > 0 {
-			return protocolError{"cannot use framing extras with non-ext packets"}
+			return invalidArgError{"cannot use framing extras with non-ext packets"}
 		}
 
 		if keyLen > math.MaxUint16 {
-			return protocolError{"key too long to encode"}
+			return invalidArgError{"key too long to encode"}
 		}
 
 		binary.BigEndian.PutUint16(headerBuf[2:], uint16(keyLen))
 	} else if pak.Magic == MagicReqExt || pak.Magic == MagicResExt {
 		if extFramesLen > math.MaxUint8 {
-			return protocolError{"framing extras too long to encode"}
+			return invalidArgError{"framing extras too long to encode"}
 		}
 
 		if keyLen > math.MaxUint8 {
-			return protocolError{"key too long to encode"}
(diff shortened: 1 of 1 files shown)


memdx/packetwriter.go
```diff
@@ -26,54 +26,54 @@ func (pw *PacketWriter) WritePacket(w io.Writer, pak *Packet) error {
 
 	if pak.Magic == MagicReq || pak.Magic == MagicRes {
 		if extFramesLen > 0 {
-			return protocolError{"cannot use framing extras with non-ext packets"}
+			return invalidArgError{"cannot use framing extras with non-ext packets"}
 		}
 
 		if keyLen > math.MaxUint16 {
-			return protocolError{"key too long to encode"}
+			return invalidArgError{"key too long to encode"}
```
programmatix, line 33: This seems reasonable

memdx/packetwriter.go
```diff
@@ -26,54 +26,54 @@ func (pw *PacketWriter) WritePacket(w io.Writer, pak *Packet) error {
...
 		}
 
 		binary.BigEndian.PutUint16(headerBuf[2:], uint16(keyLen))
 	} else if pak.Magic == MagicReqExt || pak.Magic == MagicResExt {
 		if extFramesLen > math.MaxUint8 {
-			return protocolError{"framing extras too long to encode"}
+			return invalidArgError{"framing extras too long to encode"}
```
programmatix, line 39: But this one, and most of the others here?  The client of Protostellar has no control over the framing extras, it's not exposed in the protocol at all.  If an error is hit here, is it not an internal CNG bug, rather than an InvalidArgs case?

chvck: That's true but this isn't CNG or protostellar, the callers of this memdx package aren't necessarily even gocbcorex. You'd hope any of these errors could be caught in testing well before a user had chance to see them too.

programmatix: Oh ok.  I think is yet another place where I can't really offer a useful review.  I'll +2 so you're not blocked but please get it reviewed by a competent adult at some point.

programmatix (approved):

memdx/packetwriter.go
chvck: I think that is a useful review, and it's a place where I don't think that there is strictly a correct answer. We can always selectively change it back...

PR #168, opened 2023-10-10
ING-589: Move some settings out of MutableBucketSettings

chvck:
Motivation
----------
We have some settings in MutatableBucketSettings that are not actually mutable, and should be moved to BucketSettings.


programmatix (approved):

programmatix:
Reviewing, but as ever, advise having someone who actually knows Go also review before GA.

PR #169, opened 2023-10-16
ING-598: Add handling for index exists/not found to cbqueryx

chvck:
Motivation
----------
We need to expose index exists and not found from cbqueryx, and add handling for their respective error codes. This will enable users to be able to handle these errors without having to manually parse the error that we return.

### DIFF
--- a/cbqueryx/query_test.go
+++ b/cbqueryx/query_test.go
@@ -4,9 +4,14 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/google/uuid"
 
 	"github.com/couchbase/gocbcorex/testutils"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


cbqueryx/query_test.go
```diff
@@ -45,3 +50,73 @@ func TestQuery(t *testing.T) {
...
+				"name": index,
+			},
+		},
+	})
+	body, err := json.Marshal(expectedResult)
+	require.NoError(t, err)
+
+	resp := &http.Response{
+		Status:        "success",
+		StatusCode:    200,
```
Westwooo, line 105: Why does the status and code indicate success if the index could not be found?

Westwooo (commented):
lgtm, just one question

cbqueryx/query_test.go
chvck: A good question. Query returns a 200 status code to us in this scenario, I did make sure to verify with wireshark as I asked that question myself after reviewing the gocb code.

Westwooo (approved):

PR #170, opened 2023-10-17
ING-599: Return index exists error for search index UUID mismatch

chvck:


chvck:
I could be convinced that this should be a different error, but it seems to me that it is fundamentally an already exists error.

dnault:
I believe CNG should return the same class of error it does for a CAS mismatch, since the UUID is effectively the index's CAS value.

chvck:
> I believe CNG should return the same class of error it does for a CAS mismatch, since the UUID is effectively the index's CAS value.

That was my thinking too. This is gocbcorex rather than CNG though :)

dnault:
:disappear:

PR #171, opened 2023-10-17
ING-566: Close client pools that are no longer in the cluster map

chvck:


kvclientmanager.go
```diff
@@ -159,6 +159,12 @@ func (m *kvClientManager) Reconfigure(config *KvClientManagerConfig, cb func(err
 		}
 	}
 
+	for _, pool := range oldPools {
```
Westwooo, line 162: Is this a pool of clients? Or does each pool have a single client and a pool of connections?

Westwooo (approved):
lgtm, just one question

kvclientmanager.go
chvck: So the basic breakdown here is that we have the manager which managers a collection of client pools. A client pool is specific to a node and manages a collection of clients where a single client is effectively a connection.

PR #173, opened 2023-10-19
ING-603: Use HTTP poller for cluster level agents

chvck:
Motivation
----------
At the moment we always use CCCP polling, for agents with no buckets we do not create any memcached connections. This means that those agents never receive config updates. We should use HTTP polling for those agents.


Westwooo (approved):
lgtm

PR #174, opened 2023-10-31
ING-580 Remove Query prefix from cbqueryx types

Westwooo:
Currently there are various types in cbqueryx that are prefixed with Query. This results in the types, when called outside of the package taking the form: `cbquery.Query` which is somewhat redundant.


chvck (approved):
Could you squash the 2 commits into one please?

PR #176, opened 2023-10-31
ING_622: Update scanwait and timeout to send values as go strings

chvck:


DemetrisChr:
Does the same need to be done for `kv_timeout` and `txtimeout`?

cbqueryx/query_options.go
```diff
@@ -244,7 +244,7 @@ func (o *Options) encodeToJson() (json.RawMessage, error) {
 		m["scan_vectors"] = encodeField(o.ScanVectors)
 	}
 	if o.ScanWait > 0 {
-		m["scan_wait"] = encodeField(o.ScanWait)
+		m["scan_wait"] = encodeField(o.ScanWait.String())
```
brett19, line 247: Rather than manually specifying .String() here, I think encodeField should have a time.Duration handler which does this?

brett19 (changes requested):

chvck:
> Does the same need to be done for `kv_timeout` and `txtimeout`?

Good catch, I didn't even know kvtimeout is a thing - it also looks like we were sending the field name wrong...

cbqueryx/query_options.go
chvck: Done.

cbqueryx/query_options_test.go
```diff
@@ -2,17 +2,64 @@ package cbqueryx
...
+
+	t.Run("scan_wait", func(t *testing.T) {
+		opts := &Options{
+			ScanWait: 2500 * time.Millisecond,
+		}
+
+		optsJson, err := opts.encodeToJson()
+		assert.NoError(t, err)
+
+		assert.Equal(t, `{"scan_wait":"2.5s"}`, string(optsJson))
```
chvck, line 30: I have moderate concern that the actual value here is at the whim of the Go version being used, but we can cross that bridge if the issue ever arises.

brett19 (approved):

PR #177, opened 2023-10-31
Include unit when encoding duration query options

DemetrisChr:
Currently query options which are durations are encoded as integers. The query service expects them as strings with a unit ([docs](https://docs.couchbase.com/server/current/n1ql/n1ql-rest-api/index.html#_request_parameters)), equivalent to the string representation of a Go duration


DemetrisChr:
Duplicate of #176

PR #178, opened 2023-11-01
ING-500 Add new errors to propagate to stellar-gateway

Westwooo:
In order to make the errors returned by stellar-gateway more informative we first need to make the errors returned by gocbcorex distinguishable for the cases highlighted in ING-500

### DIFF
--- a/cbmgmtx/errors.go
+++ b/cbmgmtx/errors.go
@@ -14,6 +14,7 @@ var (
 	ErrCollectionNotFound = errors.New("collection not found")
 	ErrBucketExists       = errors.New("bucket exists")
 	ErrBucketNotFound     = errors.New("bucket not found")
+	ErrServerInvalidArg   = errors.New("invalid argument")
 )
 
 type ServerError struct {
@@ -41,4 +42,17 @@ func (e contextualError) Error() string {
 
 func (e contextualError) Unwrap() error {
 	return e.Cause
+}
+
+type ServerInvalidArgError struct {
+	Argument string
+	Reason   string
+}
+
+func (e ServerInvalidArgError) Unwrap() error {
+	return ErrServerInvalidArg
+}
+
+func (e ServerInvalidArgError) Error() string {
+	return fmt.Sprintf("%s: %s - %s", e.Unwrap().Error(), e.Argument, e.Reason)
 }
--- a/cbmgmtx/mgmt.go
+++ b/cbmgmtx/mgmt.go
@@ -62,6 +62,11 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
 		}
 	}
 
+	fieldNameMap := map[string]string{
+		`durability_min_level`: "DurabilityMinLevel",
+		`replicaNumber`:        "ReplicaNumber",
+	}
+
 	var err error
 	errText := strings.ToLower(string(bodyBytes))
 
@@ -81,6 +86,12 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
 		// if we get this specific error, and its none of the above errors, then this indicates
 		// that it was the top level resource which could not be found, which is the bucket.
 		err = ErrBucketNotFound
+	} else if resp.StatusCode == 400 {
+		sErr := parseForInvalidArg(errText)
+		var ok bool
+		if sErr.Argument, ok = fieldNameMap[sErr.Argument]; ok {
+			err = sErr
+		}
 	} else if resp.StatusCode == 404 {
 		err = ErrUnsupportedFeature
 	} else if resp.StatusCode == 401 {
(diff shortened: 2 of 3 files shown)


brett19 (changes requested):

cbmgmtx/errors.go
```diff
@@ -14,6 +14,8 @@ var (
 	ErrCollectionNotFound = errors.New("collection not found")
 	ErrBucketExists       = errors.New("bucket exists")
 	ErrBucketNotFound     = errors.New("bucket not found")
+	ErrRAMQuotaTooLow     = errors.New("RAM quota too low")
```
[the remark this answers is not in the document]
chvck: We could possibly bundle all 400 status code responses that don't contain a message body that we recognise into a ErrBadRequest error. On the CNG side we *might* want to be able to differentiate some of those as failed precondition and invalid args though.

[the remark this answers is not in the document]
chvck: > Can we not do the same for the preconditionFailure errors, and if there is a 412 that we don't recognise the body of then we return an ErrPreconditionFailure?

I think some 400 status code errors could also be considered precondition failure. Note that gocbcorex doesn't have, and probably shouldn't have, a concept of precondition failure.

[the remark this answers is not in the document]
Westwooo: Sounds like we should just return `ErrBadRequest` at the moment for all 400s then if there is a specific error message that we should be indicating is a PreConditionFailure we can check the error string as with the other errors in the file

cbmgmtx/mgmt.go
```diff
@@ -81,6 +81,8 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
 		// if we get this specific error, and its none of the above errors, then this indicates
 		// that it was the top level resource which could not be found, which is the bucket.
 		err = ErrBucketNotFound
+	} else if resp.StatusCode == 400 {
```
chvck, line 84: Is this definitely correct? I can imagine that the server may respond with a 400 for more things than just invalid args?

Westwooo: I forgot to update this, IIRC @brett19 and I agreed that `BadRequest` would be more appropriate, but correct me if I'm wrong

cbmgmtx/mgmt.go
```diff
@@ -98,6 +108,33 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
...
+}
+
+func parseForInvalidArg(body string) InvalidArg {
+	var invArg InvalidArg
+	_, errs, ok := strings.Cut(body, `errors":{`)
+	if !ok {
+		return invArg
+	}
+
+	firstErr, _, ok := strings.Cut(errs, ",")
```
chvck, line 123: Is there a comma even when the server returns us only one error? Also a bit iffy that the server could return us an error message containing a command too.

cbmgmtx/mgmt.go
```diff
@@ -62,6 +62,11 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
 		}
 	}
 
+	fieldNameMap := map[string]string{
```
chvck, line 65: I love the consistency being used by the server here.

cbmgmtx/mgmt.go
```diff
@@ -98,6 +108,33 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
 	}
 }
 
+type InvalidArg struct {
```
chvck, line 111: I wonder if this should just be ServerInvalidArgError, which unwraps to ErrServerInvalidArg. Then errors.As can be used to access field and reason if we make them public instead.

cbmgmtx/mgmt.go
```diff
@@ -98,6 +108,39 @@ func (h Management) DecodeCommonError(resp *http.Response) error {
...
+	field  string
+	reason string
+}
+
+func parseForInvalidArg(body string) InvalidArg {
```
brett19, line 116: Is there some reason we can't just use json.Unmarshall for this?

brett19 (changes requested):

cbmgmtx/mgmt.go
brett19: +1

brett19 (changes requested):

cbmgmtx/mgmt.go
Westwooo: Done!

cbmgmtx/mgmt.go
Westwooo: None other than the smoothness of my brain. Changed now

cbmgmtx/errors.go
```diff
@@ -42,3 +43,17 @@ func (e contextualError) Error() string {
 func (e contextualError) Unwrap() error {
 	return e.Cause
 }
+
+type ServerInvalidArgError struct {
+	cause    error
```
chvck, line 48: I don't think this needs a cause? Unwrap should just directly always return ErrServerInvalidArg.

chvck (changes requested):

cbmgmtx/errors.go
Westwooo: Good point, done now

cbmgmtx/errors.go
```diff
@@ -42,3 +43,16 @@ func (e contextualError) Error() string {
 func (e contextualError) Unwrap() error {
 	return e.Cause
 }
+
+type ServerInvalidArgError struct {
+	argument string
```
chvck, line 48: Does users need to be able to access these fields to create their own error messages?

Westwooo: Made them public

chvck (approved):

PR #179, opened 2023-11-02
ING-605: Add watch topology config to buckets tracking manager

chvck:


brett19 (approved):

bucketstracking_agentmanager.go
```diff
@@ -263,25 +279,52 @@ func (m *BucketsTrackingAgentManager) makeAgent(ctx context.Context, bucketName
 	})
 }
 
+func (m *BucketsTrackingAgentManager) Watch(ctx context.Context) <-chan *ParsedConfig {
```
brett19, line 282: I think this should actually be `WatchConfig` since the 'Config' portion is not implied when talking about a component that is not config oriented (like `BucketsTrackingAgentManager`).  Additionally, is there some reason we couldn't just use the underlying topology provider directly?
```
func (m *BucketsTrackingAgentManager) WatchConfig(ctx context.Context) <-chan *ParsedConfig {
  return m.topologyCfgWatcher.Watch(ctx)
}
```

brett19 (changes requested):

bucketstracking_agentmanager.go
chvck: The underlying watcher is just a watcher, it doesn't filter what it pushes out based on rev id or config validity. I was thinking that by only pushing out configs applied by the manager then it dedupes that logic being scattered everywhere.

brett19: What logic is it that you're referring to?  I would expect that any deduplication logic and what not would already be getting applied by the topology watcher itself, since it's meant to be the centralized component that provides the configuration updates.

chvck: I was going to put it in the topology watcher but the topology watcher isn't actually a component, it's just a config watcher http at present. I'd feel better about creating a topology watcher which handles the deduplication logic.

topologywatcher_http.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type TopologyWatcherHttp struct {
+	logger *zap.Logger
+
+	configWatcher *ConfigWatcherHttp
+	watcherCancel func()
+}
+
+func NewTopologyWatcherHttp(config *TopologyWatcherHttpConfig, opts *TopologyWatcherHttpOptions) (*TopologyWatcherHttp, error) {
```
chvck, line 29: I'd like it more if this used DI and accepted a config watcher, but reconfigure would then have to do the same - at which point the caller has to own a config watcher and reconfigure that itself.

topologywatcher_http.go
```diff
@@ -0,0 +1,87 @@
...
+func (w *TopologyWatcherHttp) watchThread(ctx context.Context, outCh chan<- *ParsedConfig) {
+	var lastSentConfig *ParsedConfig
+	watchCh := w.configWatcher.Watch(ctx)
+
+	for {
+		var thisConfig *ParsedConfig
+		select {
+		case thisConfig = <-watchCh:
+		case <-ctx.Done():
+			close(outCh)
```
brett19, line 76: In order to guarentee the cleanup of the HTTP client (and requests), shouldn't this instead be calling w.configWatcher.Close(), and then when watchCh closes we would close outCh?

topologywatcher_http.go
brett19: What is the difference between a configWatch and a topologyWatcher?  Why can't they be one in the same?

brett19 (changes requested):

topologywatcher_http.go
chvck: They could, I just kind of liked the separation of deduplication, allowing a caller to manager it themselves if desired.

topologywatcher_http.go
chvck: There is no Close, as it's all context driven - you're right though that closing outCh should occur after watchCh closes.

agent_options.go
```diff
@@ -32,12 +35,37 @@ type AgentOptions struct {
 	HTTPConfig HTTPConfig
 }
 
+func (opts AgentOptions) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	enc.AddString("bucket-name", opts.BucketName)
```
chvck, line 39: Definitely a log redaction issue.

bucketstracking_agentmanager.go
```diff
@@ -110,13 +109,15 @@ func CreateBucketsTrackingAgentManager(ctx context.Context, opts BucketsTracking
 			tlsConfig:     opts.TLSConfig,
 			authenticator: opts.Authenticator,
 			httpTransport: httpTransport,
-			latestConfig:  bootstrapConfig,
 		},
 	}
+	m.state.latestConfig.Store(bootstrapConfig)
+
+	mgmtEndpoints := m.mgmtEndpointsLocked(bootstrapConfig)
```
Westwooo, line 116: So when we call this when we are starting CNG  `srcHTTPAddrs` will just be what we pass using `--cb-host`? Whereas `mgmtEndpoints` will be the addresses of all the nodes on the cluster gotten in the bootstrap? What's the practical implications of using `mgmtEndpoints` to create the `NewBucketsWatcherHttp` instead of  `srcHTTPAddrs`?

agent.go
```diff
@@ -248,6 +250,8 @@ func CreateAgent(ctx context.Context, opts AgentOptions) (*Agent, error) {
 		},
 	)
 
+	agent.startConfigWatcher()
```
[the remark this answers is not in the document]
chvck: The config watcher starting earlier can lead to a data race as we're assigning some of the components whilst a config is being applied.

agent_options.go
```diff
@@ -46,6 +74,15 @@ type CompressionConfig struct {
 	MinRatio             float64
 }
 
+func (c CompressionConfig) MarshalLogObject(enc zapcore.ObjectEncoder) error {
+	enc.AddBool("enabled", c.EnableCompression)
```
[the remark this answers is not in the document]
chvck: I think that's duplicated information as typically it'll be logged as a "compression-config" object, and just "enabled" seems clear to me.

bucketstracking_agentmanager.go
chvck: Possibly. Ignoring CNG srcHostAddrs is just the set of http addresses that the user provides (so maybe for CNG), mgmtEndpoints is all management endpoints on the cluster, which means that the watcher has access to use all them rather than a subset.

Westwooo: I see, also why is it called `mgmtEndpointsLocked`?

chvck: It's only safe to be called from inside the lock (once the manager has been returned from New)

brett19 (approved):

PR #181, opened 2023-11-06
ING-453 Return DocExistsErr when NotStored status returned from MutateIn

Westwooo:
MutateIn with Insert semantics can return NotStored. This should be changed so that we handle NotStored in gocbcorex for a MutateIn with Insert semantics by converting it into a DocAlreadyExists error.

### DIFF
--- a/memdx/errors.go
+++ b/memdx/errors.go
@@ -15,6 +15,7 @@ var (
 	ErrCollectionsNotEnabled               = errors.New("collections not enabled")
 	ErrDocNotFound                         = errors.New("document not found")
 	ErrDocExists                           = errors.New("document already exists")
+	ErrDocNotStored                        = errors.New("document not stored")
 	ErrValueTooLarge                       = errors.New("value too large")
 	ErrAuthError                           = errors.New("auth error")
 	ErrNotMyVbucket                        = errors.New("not my vbucket")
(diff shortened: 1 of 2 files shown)


brett19 (changes requested):
Based on more recent comments on ING-453 and ING-623, we should probably have StatusNotStored converted into ErrDocExists when the MutateIn operation is executing with Insert semantics.  In other cases, it should return ErrNotStored as it does now in this PR.

Westwooo:
> Based on more recent comments on ING-453 and ING-623, we should probably have StatusNotStored converted into ErrDocExists when the MutateIn operation is executing with Insert semantics. In other cases, it should return ErrNotStored as it does now in this PR.

@brett19 done. I haven't added a test for this due to the difficulty I had reproducing the NotStored status. Is there any way that we can mock rpcs that would allow us to manually cause a NotStoresStatus to be returned? Or is there some other method for testing suck edge cases?

brett19:
Unfortunately, this particular error shows up as a race condition if two MutateIn(w/ Insert Semantics) race together, one of them could fail with NotStored rather than DocExists.  We could theoretically add a test that validates that OpsCore correctly processes a NotStored into a DocExists, but realistically this would not be testing very much. I think you can use this test as an example if you were so inclined: https://github.com/couchbase/gocbcorex/blob/master/memdx/ops_crud_test.go#L2263

Westwooo:
> Unfortunately, this particular error shows up as a race condition if two MutateIn(w/ Insert Semantics) race together, one of them could fail with NotStored rather than DocExists. We could theoretically add a test that validates that OpsCore correctly processes a NotStored into a DocExists, but realistically this would not be testing very much. I think you can use this test as an example if you were so inclined: https://github.com/couchbase/gocbcorex/blob/master/memdx/ops_crud_test.go#L2263

I'll just leave the test then

brett19:
Sounds good.  The code looks good, just needs to be rebased on more recent changes (it won't allow me to do it in GitHub directly).

brett19 (approved):

PR #182, opened 2023-11-06
ING-624 Handle SyncWriteInProgress status

Westwooo:
Currently when we attempt to perform concurrent mutate operations with insert semantics we get an internal server error returned to the User. Instead we should handle the status and return a more useful error to STG


brett19 (approved):

brett19:
Approving, although there is another error we will need to handle that is like... SyncRecommitInProgress.

PR #183, opened 2023-11-06
ING-577: Added testing for OnBehalfOf execution.

brett19:


memdx/ops_crud_test.go
```diff
@@ -2865,6 +2865,262 @@ func TestOpsCrudValueTooLarge(t *testing.T) {
...
+
+	tests := []test{
+		{
+			Name: "Get",
+			Op: func(key []byte, opsCrud OpsCrud, cb func(interface{}, error)) (PendingOp, error) {
+				return opsCrud.Get(cli, &GetRequest{
+					Key:       key,
+					VbucketID: defaultTestVbucketID,
+					CrudRequestMeta: CrudRequestMeta{
+						OnBehalfOf: "invalid-user",
```
Westwooo, line 2892: What is the purpose of the `OnBehalfOfField`?

Westwooo (approved):
lgtm, just one question

memdx/ops_crud_test.go
brett19: OnBehalfOf allows you to authenticate as one user, but then execute an operation in the context of another.  We use it in CNG so that CNG can authenticate as one user, but then execute the operation under the context of some other authenticated user (the one from the request).

PR #186, opened 2023-11-10
ING-609, ING-610: Added helper to wait for buckets/collections.

brett19:


Westwooo (approved):
lgtm

cbmgmtx/ensurebuckethelper.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return true, nil
+}
+
+type EnsureBucketPollOptions struct {
+	Transport http.RoundTripper
+	Targets   []NodeTarget
+}
+
+func (e *EnsureBucketHelper) Poll(ctx context.Context, opts *EnsureBucketPollOptions) (bool, error) {
```
chvck, line 74: This only works for created/updated buckets but not deleted?

brett19: I think this is fine for a 'EnsureBucket' mechanism.  It's use-case is for validating that the bucket is visible to all nodes such that a asking ns_server will give a consistent answer when a memcached node returns 'bucket not found' after creation due to eventual consistency.  The inverse case where a user has delete a bucket gets this behaviour 'naturally' since we get notifications at the memcached level in the form of a dropped connection.

chvck (approved):

PR #193, opened 2023-11-18
ING-646: Handle query write statements in read-only requests.

brett19:


cbqueryx/queryrespreader.go
```diff
@@ -122,27 +125,28 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
 	if errCodeGroup == 5 {
 		err = ErrInternalServerError
 	}
-	if errCodeGroup == 12 || errCodeGroup == 14 && errCode != 12004 && errCode != 12016 {
+	if errCodeGroup == 12 || errCodeGroup == 14 {
```
Westwooo, line 128: Are these custom couchbase server error codes?

cbqueryx/queryrespreader.go
```diff
@@ -122,27 +125,28 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
 	if errCodeGroup == 5 {
 		err = ErrInternalServerError
 	}
-	if errCodeGroup == 12 || errCodeGroup == 14 && errCode != 12004 && errCode != 12016 {
```
Westwooo, line 125: Do we need to add a new if statement for  `12004` or is this alright to be `ErrIndexFailure` now?

cbqueryx/queryrespreader.go
chvck: They're codes specific to the query service, and are included as a part of the "errors" field in the response payload.

cbqueryx/queryrespreader.go
brett19: You're correct, although the existing code was actually already wrong, so I've filed an independent issue to address the fact that our 12004 cut-out wasn't actually working as it was likely intended to: https://couchbasecloud.atlassian.net/browse/ING-655

chvck (approved):

PR #196, opened 2023-11-20
ING-613: Add helpers to wait for query indexes

chvck:


cbqueryx/ensureindexhelper.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (e *EnsureIndexHelper) PollCreated(ctx context.Context, opts *EnsureIndexPollOptions) (bool, error) {
+	return e.pollAll(ctx, opts, func(numRows int) bool {
+		// If there are rows then the endpoint knows the index.
+		if numRows == 0 {
+			e.Logger.Debug("target responded with success, but the target still didn't have the index")
+			return false
+		} else if numRows > 1 {
+			e.Logger.Debug("query returned %d rows rather than 1", zap.Int("num-rows", numRows))
```
Westwooo, line 44: What does it mean if we have more than one row?

Westwooo (approved):

cbqueryx/ensureindexhelper.go
chvck: That the impossible has happened. 2 indexes with the same name on the same collection.

brett19 (approved):
I somewhat like the structure of your implementation here better than the one I used for EnsureBucket, might be worthwhile to clean up the implementations to be similar.

PR #198, opened 2023-11-21
ING-645: Improve errors returned by query index manager

Westwooo:


cbqueryx/queryrespreader.go
```diff
@@ -174,6 +181,14 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
 		err = ErrAuthenticationFailure
 	}
 
+	if errCode == 12021 {
```
brett19, line 184: I modified this file to parse errors in order of error-code so that they are easier to read and easier to find, would you be able to move these up above to be in order?

brett19 (changes requested):

brett19 (approved):

PR #203, opened 2023-11-23
ING-602: Update on demand agent manager to use a fast map

chvck:


ondemand_agentmanager.go
```diff
@@ -72,60 +77,84 @@ func (m *OnDemandAgentManager) Reconfigure(opts OnDemandAgentManagerReconfigureO
 
 // GetClusterAgent returns the Agent which is not bound to any bucket.
 func (m *OnDemandAgentManager) GetClusterAgent() (*Agent, error) {
-	m.lock.Lock()
-	defer m.lock.Unlock()
-
+	m.stateLock.Lock()
 	if m.closed {
+		defer m.stateLock.Unlock()
```
Westwooo, line 82: nit: could move this to immediately after the lock to remove the duplication

ondemand_agentmanager.go
```diff
@@ -72,60 +77,84 @@ func (m *OnDemandAgentManager) Reconfigure(opts OnDemandAgentManagerReconfigureO
...
 }
 
 // GetBucketAgent returns the Agent bound to a specific bucket, creating a new Agent as required.
 func (m *OnDemandAgentManager) GetBucketAgent(ctx context.Context, bucketName string) (*Agent, error) {
-	m.lock.Lock()
-	defer m.lock.Unlock()
-
+	m.stateLock.Lock()
 	if m.closed {
+		defer m.stateLock.Unlock()
```
Westwooo, line 94: same here

ondemand_agentmanager.go
```diff
@@ -72,60 +77,84 @@ func (m *OnDemandAgentManager) Reconfigure(opts OnDemandAgentManagerReconfigureO
...
+	agent, err := m.makeAgent(ctx, bucketName)
 	if err != nil {
 		return nil, err
 	}
+	m.slowMap[bucketName] = agent
 
-	m.bucketAgents[bucketName] = bucketAgent
+	fastMap := make(map[string]*Agent, len(m.slowMap))
+	for name, agent := range m.slowMap {
+		fastMap[name] = agent
```
Westwooo, line 130: Why are we making a new fast map, with the contents of the slow map, instead of appending the newly created agent to the existing fast map?

ondemand_agentmanager.go
brett19: I agree.  It looks like this code wasn't meant to use the defer mechanism, but rather just unlock directly, but I think I went through and mostly changed our code to use deferral instead of manual unlocking except in cases where there are coordinated locks.

ondemand_agentmanager.go
brett19: @Westwooo There are two reasons that our existing fastMaps do this (and thus I imagine Charlie followed that pattern).  One of the primary reasons is less of a technical one and more of a "defensive coding" sort of thing.  Because the fastMap critically _must_ be identical to the fastMap, we want to design the code such that it makes it unlikely that we end up with mistakes around the contents of the fastMap.  You can imagine if we add an item to the slowMap, but somehow fail to add the item to the fastMap, it means that the fastMap will be 'silently incorrect' forever.  By having the fastMap always be rebuilt it means it's extremely unlikely we could ever end up with a fastMap that's not equivalent to the slowMap.  Additionally, if we ever were to have the fastMap fall out of sync with the slowMap, the next time a modification was made to the structure, it would be 'repaired' through the rebuild process.  The second reason that this is set up this way is that because the fastMap is being atomically swapped out, it must be immutable once its placed into that atomic (or we would have race conditions around writes occuring to the fastMap while someone is trying to read it).  Since the fastMap needs to be immutable, and thus has to be a new object for each write, coupled with the first defensive coding point, it makes sense to simply build the structure from the 'source of truth' (the slowMap) each time rather than trying to copy the existing fastMap and try to make changes.

ondemand_agentmanager.go
```diff
@@ -72,60 +77,84 @@ func (m *OnDemandAgentManager) Reconfigure(opts OnDemandAgentManagerReconfigureO
...
 func (m *OnDemandAgentManager) Close() error {
-	m.lock.Lock()
-	defer m.lock.Unlock()
+	m.stateLock.Lock()
+	defer m.stateLock.Unlock()
+
+	m.opts.Logger.Debug("Closing")
+
+	m.closed = true
+	m.fastMap.Store(nil)
```
brett19, line 146: I think that we might want to 'manually' control the stateLock here (or we can continue to use defer by having some sort of 'closeAndReturnClients' function).  This would allow you to 'steal' the slow-map (and replace m.slowMap with an empty list), and then perform the closing of agents outside the lock.  Right now, calling Close() on the bucket manager can cause GetClient calls to become blocked waiting for the lock to be released (ie: blocked on all the agents shutting down).
Admittedly this is a pretty minor issue though, so I wouldn't be against merging this as-is and fixing that additional issue at some later point in time (through a separate issue, or a separate commit).

brett19 (changes requested):

ondemand_agentmanager.go
chvck: Correct, copy and paste + lack of brain issue.

ondemand_agentmanager.go
```diff
@@ -72,60 +77,94 @@ func (m *OnDemandAgentManager) Reconfigure(opts OnDemandAgentManagerReconfigureO
...
 
 	firstErr := m.clusterAgent.Close()
-	for _, agent := range m.bucketAgents {
+
+	m.slowLock.Lock()
+	var agents []*Agent
+	for _, agent := range m.slowMap {
```
brett19, line 151: nit: This is an interesting way to do `agents := m.slowMap` :D

ondemand_agentmanager.go
```diff
@@ -72,60 +77,94 @@ func (m *OnDemandAgentManager) Reconfigure(opts OnDemandAgentManagerReconfigureO
...
+}
+
+func (m *OnDemandAgentManager) getBucketAgentSlow(ctx context.Context, bucketName string) (*Agent, error) {
+	m.slowLock.Lock()
+	defer m.slowLock.Unlock()
+	if agent, ok := m.slowMap[bucketName]; ok {
+		return agent, nil
+	}
+
+	agent, err := m.makeAgent(ctx, bucketName)
```
brett19, line 122: We should file a bug to someday improve this even more and allow multiple new agents to be built simultaneously.

brett19 (approved):

ondemand_agentmanager.go
Westwooo: Makes a lot of sense, thanks for the explanation

Westwooo (dismissed):

brett19 (approved):

PR #205, opened 2023-11-27
ING-660: Add resource path to not found errors

Westwooo:


cbmgmtx/errors.go
```diff
@@ -58,3 +58,16 @@ func (e ServerInvalidArgError) Unwrap() error {
 func (e ServerInvalidArgError) Error() string {
 	return fmt.Sprintf("%s: %s - %s", e.Unwrap().Error(), e.Argument, e.Reason)
 }
+
+type ServerNotFoundError struct {
```
brett19, line 62: I would have made individual BucketNotFoundError, CollectionNotFoundError, etc... types with fields that match the specific type (for instance, a BucketName field for BucketNotFoundError).  I'm not really convinced that my approach is better though.  What was the reasoning behind your approach?

brett19: In hindsight, maybe I'm actually convinced your approach is along the lines of a better one.  Perhaps considering how Golang "stacks" error types via it's unwrap mechanism, maybe this "ServerNotFoundError" should actually be something like a "ResourceError", and thus can be reused for the "AlreadyExists" style (and others) errors as well?

brett19 (approved):

cbmgmtx/errors.go
Westwooo: My thinking was that we just wanted to maintain the same behaviour, with the ability to access the resource path if need be. Having one struct that wraps all the not found errors achieves this while declaring as few new structs as possible since we can just use errors.As to distinguish between the resource types and then just read the path. In this line of thinking I agree that we should make it more generic, then we can use if for other error types.

Westwooo: Changes this to ResourceError

cbqueryx/queryrespreader.go
```diff
@@ -155,7 +155,11 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
 		err = ErrIndexExists
 	}
 	if errCode == 12003 {
-		err = ErrCollectionNotFound
+		fields := strings.Fields(errJson.Msg)
+		err = ResourceError{
+			Cause:        ErrCollectionNotFound,
+			ResourcePath: fields[len(fields)-1],
```
chvck, line 161: I think it might be worth adding a comment somewhere here just illustrating what the msg looks like for this, so that it's clear what len(fields) - 1 is grabbing. Do we have tests somewhere for checking resource path on these errors? I'm thinking of the scenario where query changes the error message, and we want to detect that quickly.

cbmgmtx/errors.go
```diff
@@ -58,3 +58,24 @@ func (e ServerInvalidArgError) Unwrap() error {
 func (e ServerInvalidArgError) Error() string {
 	return fmt.Sprintf("%s: %s - %s", e.Unwrap().Error(), e.Argument, e.Reason)
 }
+
+type ResourceError struct {
```
chvck, line 62: Is this ever used?

chvck (approved):

PR #207, opened 2023-11-27
ING-610: Fixed naming of ManifestUID in helper.

brett19:


Westwooo (dismissed):

chvck (changes requested):
As the linter, and github, say. mgmtcomponent also needs updating, and possibly that options block too.

chvck (approved):

PR #212, opened 2023-11-30
ING-650: Return path to resource with query index errors

Westwooo:


cbmgmtx/errors.go
```diff
@@ -59,18 +59,18 @@ func (e ServerInvalidArgError) Error() string {
 	return fmt.Sprintf("%s: %s - %s", e.Unwrap().Error(), e.Argument, e.Reason)
 }
 
-type ResourceError struct {
+type ErrorWithResourceDetail struct {
```
chvck, line 62: This feels a bit like it's not in keeping with our other error types, although I don't have a suggestion for a better name...

cbqueryx/queryrespreader.go
```diff
@@ -203,7 +203,43 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
...
+
+	if err.Cause == ErrScopeNotFound || err.Cause == ErrCollectionNotFound {
+		parseResourceNotFoundMsg(&err, msg)
+	}
+
+	if err.Cause == ErrAuthenticationFailure {
+		parseAuthFailureMsg(&err, msg)
+	}
+
+	if errors.Is(err.Cause, ErrServerInvalidArg) {
```
chvck, line 219: Why do we use a mix of == and .Is?

Westwooo: Because with the InvalidArg error the cause is `ErrInternalServerError` which in turn is caused by the `InvalidArgErr` so we need to use Is here to unwrap down to the InvalidArgErr, whereas the other cases don't have this error nesting

cbmgmtx/errors.go
Westwooo: Thoughts on the new name?

cbqueryx/queryrespreader.go
Westwooo: I've changed them all to errors.Is for consistency and robustness to future wrapping

cbqueryx/queryrespreader.go
```diff
@@ -214,28 +250,89 @@ func createResourceError(msg string, cause error) ResourceError {
...
-	}
 	_, trimmedPath, found := strings.Cut(path, ":")
 	if !found {
-		return err
+		return
 	}
-
 	fields = strings.Split(trimmedPath, ".")
-	if cause == ErrScopeNotFound {
+	if err.Cause == ErrScopeNotFound {
```
chvck, line 258: .Is here and below too?

cbqueryx/queryrespreader.go
```diff
@@ -204,6 +204,42 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
 }
 
 func createResourceError(msg string, cause error) ResourceError {
+	err := ResourceError{
+		Cause: cause,
+	}
+
+	if errors.Is(err, ErrScopeNotFound) || errors.Is(err, ErrCollectionNotFound) {
+		parseResourceNotFoundMsg(&err, msg)
```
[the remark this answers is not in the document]
Westwooo: I think it's negligible, but I did write it so I'm a bit biased. The main reason I did it like this was so that we could have
```
err := ResourceError{
		Cause: cause,
	}
``` 
in just one place, if they return the error, I think they'd need to take the cause and construct it themselves. But this isn't a crucial reason so if it read awkwardly I can change it.

chvck (approved):

PR #213, opened 2023-12-01
ING-523: Improve GetAllReplicas handling

Westwooo:


agent_ops.go
```diff
@@ -20,6 +21,16 @@ func (agent *Agent) GetReplica(ctx context.Context, opts *GetReplicaOptions) (*G
 	return agent.crud.GetReplica(ctx, opts)
 }
 
+func (agent *Agent) GetAllReplicas(ctx context.Context, opts *GetAllReplicasOptions) ([]GetAllReplicasResult, error) {
+	bucket, err := agent.mgmt.GetBucket(ctx, &cbmgmtx.GetBucketOptions{
```
brett19, line 25: This should not be neccessary.  If you take a look at how CrudComponent executes a GetReplica operation, it's execution stack eventually invokes OrchestrateMemdRouting, which is responsible for identifying the place that a particular operation is being sent.  Additionally, this implementation would still have the same problem mentioned in the ING where a change to the servers for a replica could cause GetAllReplicas to miss a replica or duplicate the result from a specific replica.  What we should instead do is use the same topological information that OrchestrateMemdRouting uses to identify which nodes to send an operation to in a 'live' manner, and additionally the logic should be adjusted such that we do not refer to the replicas we are trying to fetch by their arbitrary 'replicaIdx' but rather by the server that the request is being sent to.  We already deal with this a bit in our handling of routing, where we route operations to a nodeId rather than to a specific vbucket number, such that we can maintain the 'atomicity' of this information.

brett19 (changes requested):

agent_ops.go
Westwooo: Yeah I thought I was missing the last thing. Have made some changes

Westwooo (commented):
WIP: needs some tidying up, also need to finalise some details about how to handle changing of orchestrator

crud.go
```diff
@@ -152,6 +153,203 @@ func (cc *CrudComponent) GetReplica(ctx context.Context, opts *GetReplicaOptions
...
+										}
+
+										return results, nil
+									}
+
+									// Here we know that len(results) == i == num_replicas
+									// So we have fetched a sufficient number of replicas, now we just need to get the original
+									_, err = routingFn(0, collectionID)
+									if err != nil {
+										result.OutCh <- &ReplicaStreamEntry{
```
Westwooo, line 318: Are we are okay returning a replica that no longer exists on that node? Say we get the original from node1 and then a replica from node2. Now a rebalance occurs and node2 becomes the master while node3 has the replica. We'd then fetch the new original from node2 and return. So the user would receive the replica that was initially on node2 and then the master from node2.

brett19: I'm going to put some more analysis of these situations into the ING for later reference on our thought process as this appears that it's a lot more complicated than it appeared on the surface.

PR #217, opened 2023-12-05
ING-676: Added helpers for ensuring creation of search indexes.

brett19:


brett19:
I've intentionally re-run our test pipeline on this PR a number of times and it has never reproduced the transient test failure we see occasionally, so hopefully this should resolve that.

Westwooo (approved):

PR #225, opened 2023-12-06
ING-688: Extend search index timeout.

brett19:


thejcfactor (approved):
Just a nit-pick and it doesn't impact the change, but maybe a variable for the timeout? Makes a future change a one-liner 😄

brett19:
Ideally these tests shouldn't need to be this long, and hopefully we never need to change them again.  Realistically each test should have it's own timeout anyways.

PR #228, opened 2023-12-07
ING-690: Ensure the search index on all nodes during testing.

brett19:


avsej (approved):
I didn't know about /api/cfgRefresh. Looks good

PR #230, opened 2023-12-11
ING-681: Add support for query index management

chvck:


chvck:
7.0.3 doesn't support IF (NOT) EXISTS syntax...

brett19:
Oops.  We will need to switch to using normal queries and handling the error instead :S

Westwooo (approved):

Westwooo (approved):

PR #234, opened 2023-12-18
ING-652: Add support for bucket history settings

Westwooo:


cbmgmtx/mgmt.go
```diff
@@ -623,6 +639,14 @@ func (h Management) decodeMutableBucketSettings(data *bucketSettingsJson) (*Muta
 	settings.MaxTTL = time.Duration(data.MaxTTL) * time.Second
 	settings.CompressionMode = CompressionMode(data.CompressionMode)
 	settings.DurabilityMinLevel = DurabilityLevel(data.MinimumDurabilityLevel)
+	settings.HistoryRetentionBytes = data.HistoryRetentionBytes
+	settings.HistoryRetentionSeconds = data.HistoryRetentionSeconds
+
+	if data.StorageBackend != "magma" {
+		settings.HistoryRetentionCollectionDefault = nil
```
chvck, line 646: This seems like it should be rejected or passed through to be handled server side.

Westwooo: ns_server always returns a bool value for this field, no matter what the storage type is. So I wanted to differentiate between false and the zero value.

chvck: I think we should just return the bool that ns_server does, *but* possibly making it a bool pointer in the JSON as older server versions certainly won't return this field at all.

Westwooo: Done

chvck (approved):

PR #235, opened 2023-12-20
Limit tests in CI to only run a single package at a time

chvck:
Tests can randomly fail due to our usage of cbdinocluster impacting tests within different packages running in parallel.


brett19 (approved):
Looks good, although we should rerun the tests against this commit a few times to confirm it actually made an improvement.

PR #238, opened 2023-12-22
ING-715: More useful FTS errors

Westwooo:
There are a few FTS errors that are not parsed by gocbcorex, resulting in an unknown error status being returned to the User. This PR addresses such cases.

### DIFF
--- a/cbsearchx/search.go
+++ b/cbsearchx/search.go
@@ -116,6 +116,9 @@ func (h Search) UpsertIndex(
 	if opts.Type == "" {
 		return errors.New("must specify index type when creating an index")
 	}
+	if opts.SourceType == "" {
+		return errors.New("must specify source type when creating an index")
+	}
 
 	var reqURI string
 	if opts.ScopeName == "" && opts.BucketName == "" {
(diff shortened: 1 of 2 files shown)


cbsearchx/search.go
```diff
@@ -586,6 +589,13 @@ func (h Search) DecodeCommonError(resp *http.Response) error {
 		err = ErrIndexExists
 	} else if strings.Contains(errText, "current index uuid") && strings.Contains(errText, "did not match input uuid") {
 		err = ErrIndexExists
+	} else if strings.Contains(errText, "unknown indextype") {
+		err = ErrUnknownIndexType
+	} else if strings.Contains(errText, "error obtaining vbucket count for bucket") ||
+		strings.Contains(errText, "requested resource not found") {
```
chvck, line 595: I suspect this can be returned to us for other cases too?

Westwooo: I'll add some more thorough tests but I don't think that it can. Only a missing source returns this error, since if we specify a Bucket/Scope name in the index definition that doesn't exist we get ErrIndexNotFound since the bucket and scope name are only used to construct the fully qualified index name.

Westwooo: I can't find any other cases where this message is returned, the tests are in ING-677

Westwooo: The error you get is: `search service not available`

chvck (approved):

PR #239, opened 2024-01-04
ING-677 cleanup search tests

Westwooo:


cbsearchx/ensureindexhelper_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestEnsureIndex(t *testing.T) {
+func TestIndexPolling(t *testing.T) {
```
brett19, line 16: We generally have been calling these kinds of tests "Ensure" tests.  Any reason for the divergence here?

brett19 (approved):

cbsearchx/ensureindexhelper_test.go
Westwooo: Because I added a test for `EnsureIndex` in searchcomponent_test.go` which made sense to be called `EnsureIndexTest` and I didn't want two tests with the same name. But I'm happy to change the names if it goes against the current naming scheme.

PR #243, opened 2024-01-15
Improve cbauth integration

brett19:


cbauthx/cbauth.go
```diff
@@ -0,0 +1,437 @@
...
+	password string,
+	clusterUuid string,
+) (*CbAuthClient, string, error) {
+	// Do not attempt to connect to cbauth faster than allowed
+	{
+		lastConnectDelta := time.Since(a.lastConnectTime)
+		if lastConnectDelta < a.minReconnectTime {
+			select {
+			case <-time.After(a.minReconnectTime - lastConnectDelta):
+			case <-ctx.Done():
```
chvck, line 165: Should this do something other than just keep running?

cbauthx/cbauth.go
```diff
@@ -0,0 +1,437 @@
...
+		a.lastConnectTime = time.Now()
+	}
+
+	var availableEndpoints []string
+	var failedEndpoints []string
+	var failedErrors []error
+	for {
+		availableEndpoints = availableEndpoints[:0]
+
+		a.lock.Lock()
```
chvck, line 178: Why is there a new to lock here?

cbauthx/cbauth.go
```diff
@@ -0,0 +1,437 @@
...
+			lock.Unlock()
+
+			if a.closed {
+				break
+			}
+		}
+	}()
+
+	for {
+		if initialCli != nil {
```
chvck, line 296: Confused what this is doing, surely cli is already assigned to initialCli if initialCli isn't nil?

cbauthx/cbauth.go
```diff
@@ -0,0 +1,437 @@
...
+
+func (a *CbAuth) CheckUserPass(ctx context.Context, username string, password string) (UserInfo, error) {
+	cli := a.currentClient.Load()
+	info, err := cli.CheckUserPass(ctx, username, password)
+	if err != nil {
+		if errors.Is(err, ErrInvalidAuth) {
+			return UserInfo{}, ErrInvalidAuth
+		}
+
+		a.logger.Debug("failed to check user with cbauth", zap.Error(err))
```
chvck, line 374: Is this technically request level logging?

cbauthx/cbauthclient.go
```diff
@@ -0,0 +1,281 @@
...
+	return cli, nil
+}
+
+func (c *CbAuthClient) refreshLivenessLocked() {
+	// store the latest time for the heartbeat
+	c.lastCommTs.Store(time.Now().Unix())
+
+	// Reset our timer so we don't close the connection
+	if c.heartbeatTimer != nil {
+		c.heartbeatTimer.Reset(c.livenessTimeout)
```
chvck, line 176: It's slightly confusing that this is a different timeout to just above.

cbauthx/cbauth.go
brett19: Yes.  Good point.

cbauthx/cbauth.go
brett19: A mistake from when I moved the endpoints to being read from inside the function, to being passed in.

cbauthx/cbauth.go
brett19: This is true, but I wanted the logic to be more consolidated.  I'll adjust the variable declaration to no longer default to initialCli though.

cbauthx/cbauth.go
brett19: It's at debug level, so should be fine.

cbauthx/revrpcclient_test.go
```diff
@@ -0,0 +1,38 @@
+package cbauthx
+
+/*
```
chvck, line 3: hmm

brett19: Ahh, forgot to remove this from the commit.

chvck (approved):

PR #245, opened 2024-01-17
ING-735, ING-736: CbAuth racing and reconnecting

brett19:


cbauthx/authcheck_http.go
```diff
@@ -80,6 +81,17 @@ func (a *AuthCheckHttp) CheckUserPass(ctx context.Context, username string, pass
 		}
 	}
 
+	// In the interest of mitigating the impact of an incorrect auth check url
+	// being used, we validate some expected state of the response which has the
+	// effect of significantly reducing the chances we accept auth from those
+	// urls which do not actually check auth.
```
Westwooo, line 87: What is the impact of using an incorrect auth check URL that is being mitigated here?

cbauthx/cbauth.go
```diff
@@ -266,7 +266,9 @@ func (a *CbAuth) Reconfigure(config *CbAuthConfig) error {
 	}
 
 	if !a.isCliValidLocked() {
-		a.cli.Close()
+		if a.cli != nil {
```
Westwooo, line 269: These changes are in both PRs, not sure which one they belong in

Westwooo (approved):

cbauthx/authcheck_http.go
brett19: Imagine if for some reason, this was initialized with an authCheckURL of "", when we concatenate the source server url of "http://localhost:8091" to this, we would accidentally be using the index-page for Couchbase server rather than the _cbauth url.  In this case, by checking that the response properties match our expectation, it ensures we are unlikely to accept a pages response unless it was a cbauth response.

cbauthx/cbauth.go
brett19: This change's base-branch is the other one, once it's merged and this PR rebased, those changes should disappear from here.

cbauthx/authcheck_http.go
Westwooo: I see, thanks for the exaplination

PR #246, opened 2024-01-17
ING-710: Add cbauthx additional testing.

brett19:


cbauthx/cbauth_test.go
```diff
@@ -108,3 +112,228 @@ func TestCbAuthBasicSlow(t *testing.T) {
...
+	blockHost := blockNode.Hostname
+	blockNsEndpoint := blockNode.NsEndpoint()
+
+	var allEndpoints []string
+	for _, node := range nodes {
+		allEndpoints = append(allEndpoints, node.NsEndpoint())
+	}
+
+	log.Printf("orchestrator: %s", orchEndpoint)
+	log.Printf("nodes:")
```
Westwooo, line 243: I assume these printfs were left here deliberately?

cbmgmtx/mgmt.go
```diff
@@ -1016,3 +1018,164 @@ func (h Management) FlushBucket(
...
+	if opts.Domain == "" {
+		opts.Domain = "local"
+	}
+
+	posts := url.Values{}
+
+	if opts.DisplayName != "" {
+		posts.Add("name", opts.DisplayName)
+	}
+	if opts.Password != "" {
```
Westwooo, line 1121: Is it valid to create a user without a password?

cbmgmtx/mgmt.go
```diff
@@ -1016,3 +1018,164 @@ func (h Management) FlushBucket(
...
+	Username   string
+	OnBehalfOf *cbhttpx.OnBehalfOfInfo
+}
+
+func (h Management) DeleteUser(
+	ctx context.Context,
+	opts *DeleteUserOptions,
+) error {
+	if opts.Username == "" {
+		return errors.New("must specify username when upserting a user")
```
Westwooo, line 1159:
```suggestion
		return errors.New("must specify username when deleting a user")
```

Westwooo (dismissed):

cbauthx/cbauth_test.go
brett19: Yea, they help when debugging after the test fails.

cbmgmtx/mgmt.go
brett19: I'm not entirely sure, but I didn't think there was much harm in avoiding sending the field if the field is blank anyways.  I suspect you're underlying assumption is probably correct though, and the server would reject the request.

cbmgmtx/mgmt.go
brett19: Good catch.

Westwooo (approved):

PR #248, opened 2024-01-19
ING-744: Disable auto-failover during dino-tests.

brett19:


testutilsint/dinocluster.go
```diff
@@ -85,6 +94,32 @@ func (c *DinoController) cleanup() {
...
+	require.NoError(c.t, err)
+}
+
+func (c *DinoController) EnableAutoFailover() {
+	if c.oldFoSettings == nil {
+		return
+	}
+
+	err := getTestMgmt().ConfigureAutoFailover(context.Background(), &cbmgmtx.ConfigureAutoFailoverRequest{
+		Enabled: ptr.To(c.oldFoSettings.Enabled),
```
Westwooo, line 118: Why not just have this as `Enabled: ptr.to(true)`?

brett19: A bit contrary to the name of the function, this is actually restoring the cluster to whatever state it was in before the test started.  It's exposed as a public function, but it's invoked automatically as part of test cleanup through the hook installed when the DinoController is constructed above.

Westwooo (approved):

PR #257, opened 2024-03-01
ING-523: Improve GetAllReplicas handling

Westwooo:
The logic here starts a thread for each replica that should exist according to NUM_REPLICAS. The overall GetAllReplicas operation is complete when NUM_REPLICAS + 1 docs have been returned, or it hits the timeout. The individual threads will run indefinitely until these points, even if the thread has returned a document. This is because the individual requests are routed based on the replica ID, which can be routed to a different node if there is a rebalance during the GetAllReplicas.

### DIFF
--- a/crud.go
+++ b/crud.go
@@ -2,9 +2,14 @@ package gocbcorex
 
 import (
 	"context"
+	"errors"
+	"sync"
+	"time"
 
 	"github.com/couchbase/gocbcorex/memdx"
+	"go.uber.org/atomic"
 	"go.uber.org/zap"
+	"golang.org/x/exp/slices"
 )
 
 type CrudComponent struct {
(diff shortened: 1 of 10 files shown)


retrymanager_getallreplicas.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func (m *RetryManagerGetAllReplicas) NewRetryController() RetryController {
+	return &retryControllerGetAllReplicas{
+		parent: m,
+	}
+}
+
+type retryControllerGetAllReplicas struct {
```
brett19, line 26: Why do we need a separate retry controller for GetAllReplicas?

crud_int_test.go
```diff
@@ -0,0 +1,95 @@
...
+		if !result.IsReplica {
+			masterCount += 1
+		} else {
+			replicaCount += 1
+		}
+		result, err = stream.Next()
+	}
+	require.Equal(t, NUM_REPLICAS, masterCount+replicaCount)
+
+	testutilsint.SkipIfNoDinoCluster(t)
```
brett19, line 63: This should be at the top of the test.  If we are doing two "types" of tests (one dino, one not), we should have two tests, one that uses dino and one that doesnt.

crud_int_test.go
```diff
@@ -0,0 +1,95 @@
...
+	nodes := testutilsint.GetTestNodes(t)
+	nonOrchestrators := nodes.Select(func(node *testutilsint.NodeTarget) bool {
+		return !node.IsOrchestrator
+	})
+
+	dino := testutilsint.StartDinoTesting(t, true)
+	for _, node := range nonOrchestrators {
+		dino.BlockAllTraffic(node.Hostname)
```
brett19, line 71: In this line, you're blocking all non-orchestrators from talking to eachother, but you've already previously run the Upsert and then the GetAllReplicas.  This means it's likely that the document will already have propagated to all the nodes, but more importantly it means that the following operation should be guarenteed to fail since it will be unable to contact the majority of the nodes.  A more representative test might be to block one node that is a replica for a document, perform an upsert to the primary node, and then GetAllReplicas.  We would expect that the GetAllReplicas operation would fail until you've unblocked the final node from the replicas.

crud.go
```diff
@@ -152,6 +157,230 @@ func (cc *CrudComponent) GetReplica(ctx context.Context, opts *GetReplicaOptions
...
+										return res, err
+									}
+									mu.Unlock()
+									return nil, err
+								})
+
+							// The replicaID is being routed to a node from which we have already received a result
+							// therefore we wait and then try again, incase the replicaIds have been changed by a rebalance
+							if res == nil && err == nil {
+								time.Sleep(10 * time.Millisecond)
```
brett19, line 314: Rather than a hardcoded wait here, shouldn't this be based on the retry manager in some way?

retrymanager_getallreplicas.go
Westwooo: I did this because for the individual replica reads we want to retry for `ErrNoServerAssigned` as this is what we get during a rebalance is ongoing, and the normal retry manager fails on this error.

crud_int_test.go
Westwooo: Done!

crud_int_test.go
Westwooo: With this I wanted the replica to already be present on these nodes, but I wanted them to be unreachable when we run the GetAllReplicas. I wouldn't expect the command to fail I'd just expect the single document from the orchestrator to be returned. I wasn't aware that we wanted the get all Replicas to fail on unreachable nodes, I thought the plan was to return all available replicas, allowing the user to continue in the case where they are using this to get any available replica.

crud.go
Westwooo: Done!

retrymanager_getallreplicas.go
brett19: In fairness, we should probably retry this error generally I think.

crud.go
```diff
@@ -152,6 +157,230 @@ func (cc *CrudComponent) GetReplica(ctx context.Context, opts *GetReplicaOptions
...
+		}, nil
+	}
+
+	var endpoints []string
+	var mu sync.Mutex
+	var sendLock sync.Mutex
+	var returnedResults uint32
+	var numReplicas atomic.Uint32
+
+	temp, err := cc.vbs.NumReplicas()
```
brett19, line 263: We should try and avoid calling things generic names like `temp`.  Instead, you can probably call it `initialNumReplicas` or something along those lines.

brett19 (approved):
I've put a few comments on here, but Approved it as they are minor items that could be fixed in a future PR.  Additionally, it may be worthwhile to write a `docs/REPLICA_READ.md` file that describes the overarching behaviour that you've implemented here, since there was a lot of thinking that went into how this would work.

crud.go
Westwooo: Done!

retrymanager_getallreplicas.go
brett19: I think that by the definition of how GetReplicas works, I don't think it ever makes sense to return duplicates from the same node, so it probably makes sense to collapse that behaviour into the replica code itself, rather than being part of the retry manager.  Additionally, the `ErrRepeatedReplicaRead` error only appears during GetReplicas anyways, so having it in the default retry path wouldn't have any impacts.  There is already precedence for having operation-specific errors be exposed as unique errors for the purposes of having a generalized retry system as well.

brett19 (approved):

PR #258, opened 2024-03-05
ING-763: Authoritative Not Found

brett19:


retrymanager_consistency.go
```diff
@@ -0,0 +1,122 @@
...
+					memdResErr.ScopeName)
+				if existsErr != nil {
+					return 0, false, &contextualError{
+						Message: "failed to double-check memd whether scope exists",
+						Cause:   existsErr,
+					}
+				}
+
+				if exists {
+					return 500 * time.Millisecond, true, nil
```
chvck, line 45: Are these meant to always be a fixed 500ms?

bucketchecker.go
```diff
@@ -0,0 +1,7 @@
+package gocbcorex
+
+import "context"
+
+type BucketChecker interface {
```
chvck, line 5: We have BucketChecker vs ManifestFetcher, should probably make these consistent?

chvck: I guess they are subtly different in behaviour.

retrymanager_consistency.go
brett19: It is, since the timeline for these cases happening doesn't really benefit from exponential decay.

retrymanager_default.go
```diff
@@ -40,9 +41,9 @@ func (rc *retryControllerDefault) isRetriableError(err error) bool {
 		errors.Is(err, ErrCollectionManifestOutdated)
 }
 
-func (rc *retryControllerDefault) ShouldRetry(err error) (time.Duration, bool) {
+func (rc *retryControllerDefault) ShouldRetry(ctx context.Context, err error) (time.Duration, bool, error) {
```
Westwooo, line 44: Why do we need the added return error here if it's always nil?

brett19: It's nil in this case, but it's not nil in the other retry controller for consistency.

Westwooo (approved):

PR #261, opened 2024-03-26
ING-790: Added connection error throttling and improved logging.

brett19:
This adds throttling for connecting clients in a kvClientPool such that quickly errors do not lead to spamming connection attempts.  We also updated the logging around client management to make it easier to diagnose faults related to connections in the future.

### DIFF
--- a/kvclientpool.go
+++ b/kvclientpool.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/google/uuid"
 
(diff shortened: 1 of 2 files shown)


kvclientpool.go
```diff
@@ -196,12 +198,37 @@ func (p *kvClientPool) startNewClientLocked() <-chan struct{} {
...
-		p.logger.Debug("creating new client")
+		p.logger.Info("creating new client kv client",
+			zap.Any("clientConfig", clientConfig))
+
+		p.lock.Lock()
+		defer p.lock.Unlock()
+
+		connectErrThrottle := 1 * time.Second
+		for {
+			connectWaitPeriod := connectErrThrottle - time.Since(p.connectErrTime)
```
chvck, line 209: Doesn't this cause a delay on initial connect?

kvclientpool.go
```diff
@@ -196,12 +198,37 @@ func (p *kvClientPool) startNewClientLocked() <-chan struct{} {
...
+			if connectWaitPeriod > 0 {
```
chvck, line 210: If we invert this too if == 0 { break} then we can unnest most of the code and remove the continue call?

kvclientpool.go
brett19: `connectErrTime` would be time.Time{} (0) on first connect, which yields a huge negative number and fails the next check.

kvclientpool.go
brett19: When dealing with loops, I usually try to keep the 'common path' to be the one that consumes the body, and then have the 'exceptional case' be in an if block, in this case, the standard case would be that we break without doing anything.

chvck (approved):

PR #267, opened 2024-04-09
Allow MaxTTL to have a negative value

Westwooo:


brett19 (approved):
Just to double check, we decided this should be a int32 rather than int64?

PR #273, opened 2024-05-08
ING-663: Added support for fts vector search.

brett19:
This commit also adds some mechanisms to enable cluster capabilities to be propagated to our components.


Westwooo (approved):
lgtm!

chvck (commented):
Looks good but think we could do with some tests

chvck (approved):

PR #274, opened 2024-05-20
ING-799: Consolidate default and getAllReplicas retry managers

Westwooo:


brett19:
I suspect this flakiness in the test might be caused by the same thing which was causing our durability tests to fail.  In some cases, the dino tests run before these particular tests, and they put the cluster in a state where replication links are 'damaged' and not working.  This causes the tests to fail because the expected replication fails to occur (the other ticket is a case of sync durability failing).

retrymanager_default.go
```diff
@@ -37,7 +37,9 @@ func (rc *retryControllerDefault) isRetriableError(err error) bool {
 		errors.Is(err, memdx.ErrSyncWriteReCommitInProgress) ||
 		errors.Is(err, cbhttpx.ErrConnectError) ||
 		errors.Is(err, ErrVbucketMapOutdated) ||
-		errors.Is(err, ErrCollectionManifestOutdated)
+		errors.Is(err, ErrCollectionManifestOutdated) ||
+		errors.Is(err, ErrNoServerAssigned) ||
+		errors.Is(err, ErrRepeatedReplicaRead)
```
brett19, line 42: I think this ErrRepeatedReplicaRead should probably be handled directly by the error handling in GetReplicas rather than being pushed up to the retry handler.  Why would we be retrying in this case?

Westwooo: Because even when we have already read a replica, we want to keep trying in that replicaIds thread. Since if there is a rebalance then the same replicaID could be routed to a replica on a node that we have not yet read.

brett19 (approved):

brett19 (approved):

PR #281, opened 2024-07-02
Transactionx fix ups

brett19:


transactionsx/constants.go
```diff
@@ -55,26 +55,25 @@ func (state TransactionAttemptState) String() string {
...
 // NOTE: The errors within this section are critically ordered, as the order of
 // precedence used when merging errors together is based on this.
 const (
 	// TransactionErrorReasonSuccess indicates the transaction succeeded and did not fail.
-	TransactionErrorReasonSuccess TransactionErrorReason = iota
+	TransactionErrorReasonSuccess TransactionErrorReason = TransactionErrorReason(0)
 
 	// TransactionErrorReasonTransactionFailed indicates the transaction should be failed because it failed.
-	TransactionErrorReasonTransactionFailed
+	TransactionErrorReasonTransactionFailed = TransactionErrorReason(1)
```
chvck, line 67: Should all these not have the Transaction prefix?

transactionsx/hooks.go
```diff
@@ -45,7 +45,6 @@ type TransactionHooks interface {
 }
 
 // TransactionCleanupHooks provides a number of internal hooks used for testing.
-// Internal: This should never be used and is not supported.
 type TransactionCleanupHooks interface {
```
chvck, line 48: Also the transaction prefix here.

chvck (approved):

transactionsx/constants.go
brett19: I left the Transaction prefix on things that are specific to the Transaction object.

PR #283, opened 2024-07-05
ING-827: Added support for pinning queries to an endpoint.

brett19:


agentmanager_int_test.go
```diff
@@ -66,7 +67,9 @@ func TestOnDemandAgentManagerClose(t *testing.T) {
 	require.NoError(t, err)
 
 	_, err = agent.Query(context.Background(), &gocbcorex.QueryOptions{
-		Statement: "SELECT 1=1",
+		QueryOptions: cbqueryx.QueryOptions{
```
chvck, line 70: This feels a bit odd.

brett19: The main distinction for why this happened is that at the cbqueryx level (where the initial QueryOptions are defined) don't really deal with the routing aspect of the requests.  They accept a single endpoint and then perform queries against them.  At the gocbcorex level, where we actually perform routing and what not is where endpoints are relevant.  We previously had simply aliased cbqueryx.QueryOptions to gocbcorex.QueryOptions, however with these new routing-specific options, I needed to split them apart (similar to what we do with the memdx operations).  As opposed to the memdx sort of operations though, where some the options no longer make sense and should be hidden from the top-level interface (which is why memdx copies all the relevant fields to a new struct), in query all the lower-level interface options are still valid and so I opted to simply "inherit" the cbsearchx.QueryOptions in gocbcorex.QueryOptions.  There is however an alternative approach though, where gocbcorex.QueryOptions is a distinct structure from cbsearchx.QueryOptions, and we simply copy the fields over to build the lower-level request.

chvck: I'm a bit torn on this one. It seems possible that in the future we could add more gocbcorex level query options that aren't applicable at the queryx level though, so maybe the approach of splitting them makes sense.

Westwooo (dismissed):
LGTM

agentmanager_int_test.go
brett19: I've adjusted the gocbcorex.QueryOptions to match what we do with memdx and simply duplicate the fields, and copy them into the underlying request.

chvck (approved):

PR #293, opened 2024-10-29
ING-959: Refactored hostname handling when parsing configs.

brett19:
This commit also refactors how we deal with remote hostnames and addresses with the various clients and connections inside of the library.  Primarily with the intent of simplifying the interfaces that we expose, and ensuring the correct details are available in the right places.

### DIFF
--- a/kvclientmanager.go
+++ b/kvclientmanager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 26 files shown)


kvclientmanager.go
```diff
@@ -285,6 +286,22 @@ func (m *kvClientManager) Close() error {
 	return nil
 }
 
+type KvClientError struct {
```
chvck, line 289: Aren't our errors typically in the errors file?

kvclientpool.go
```diff
@@ -505,24 +504,24 @@ func (p *kvClientPool) Reconfigure(config *KvClientPoolConfig, cb func(error)) e
 
 		// reconfiguring is successful up until this point, so it can stay in the
 		// current list of clients.  it may be moved later by the Reconfigure callback.
-		numReconfigured++
+
+		// if the pool name changed as part of this reconfigure, move it.
+		if oldPoolName != newPoolName {
+			p.connCountMetric.Record(context.Background(), -1, metric.WithAttributes(
```
chvck, line 510: This is now in the loop so is going to record the metrics for every client that gets reconfigured, is that correct?

chvck (approved):

kvclientpool.go
brett19: Yea, but we only record -1/+1 for each client that is reconfigured rather than doing them all with multiple-counts like it had previously.

kvclientmanager.go
brett19: Not always.  Admittedly, it's actually a practice I'm not entirely fond of, but admittedly I also don't like having them scattered _everywhere_ either... Hmm.

PR #301, opened 2024-12-13
ING-989: Fixed memd status code 0x6 to be DeltaBadval.

brett19:
We previously have always called this BadDelta, but this name is not accurate for what this error is actually used for from memcached.

### DIFF
--- a/memdx/status.go
+++ b/memdx/status.go
@@ -24,8 +24,8 @@ const (
 	// StatusNotStored occurs when the server fails to store a key.
 	StatusNotStored = Status(0x05)
 
-	// StatusBadDelta occurs when an invalid delta value is specified to a counter operation.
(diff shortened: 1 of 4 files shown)


memdx/status.go
```diff
@@ -24,8 +24,8 @@ const (
 	// StatusNotStored occurs when the server fails to store a key.
 	StatusNotStored = Status(0x05)
 
-	// StatusBadDelta occurs when an invalid delta value is specified to a counter operation.
-	StatusBadDelta = Status(0x06)
+	// StatusDeltaBadval occurs when an invalid delta value is specified to a counter operation.
```
avsej, line 27:
```suggestion
	// StatusDeltaBadval occurs when an document value cannot be used as a counter.
```

avsej (approved):

PR #302, opened 2024-12-17
ING-782: Add handling for query index build fails

chvck:


cbqueryx/queryrespreader.go
```diff
@@ -136,6 +136,9 @@ func (r *queryRespReader) parseError(errJson *queryErrorJson) *ServerError {
 		if strings.Contains(lowerMsg, "build already in progress") {
 			err = ErrBuildAlreadyInProgress
 		}
+		if strings.Contains(lowerMsg, "build index fails") {
```
brett19, line 139: This message should probably include the 'will retry' text, since that's an important detail to how this error is used.

brett19 (changes requested):

Westwooo (approved):

brett19 (approved):

PR #304, opened 2025-01-20
ING-996: Lower processing overheads

brett19:


memdx/ops_crud.go
```diff
@@ -166,7 +166,8 @@ type GetResponse struct {
 }
 
 func (o OpsCrud) Get(d Dispatcher, req *GetRequest, cb func(*GetResponse, error)) (PendingOp, error) {
-	reqMagic, extFramesBuf, err := o.encodeReqExtFrames(req.OnBehalfOf, 0, 0, false, nil)
+	extFramesBuf := make([]byte, 0, 128)
```
Westwooo, line 169: What's the reason for declaring this and passing it into `encodeReqExtFrames` vs just passing in nil as was done before?

Westwooo (commented):
Nice change, mostly LGTM, just a nit and a question

memdx/ops_crud.go
chvck: At a guess it preallocates the capacity so it doesn't need to grow the array?

kvclienttelem.go
```diff
@@ -0,0 +1,146 @@
...
+
+	startTime  time.Time
+	bucketName string
+	opName     string
+	span       trace.Span
+}
+
+func (k *kvClientTelem) BeginOp(
+	ctx context.Context,
+	c *kvClient,
```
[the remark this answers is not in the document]
brett19: It gets used from kvclient.go

memdx/ops_crud.go
brett19: Go has an optimization where if you allocate an array with a constant size (up to some limit) and the compiler is able to prove that the allocated memory does not need to escape from the heap, it actually allocates on the heap instead which is what occurs here.  This `make([]byte, 0, 128)` call doesn't actually perform an allocation.

Westwooo: I see, thanks Brett

chvck (approved):

PR #319, opened 2025-07-04
ING-1115: Ignore not found errs from replica reads

Westwooo:


brett19:
Looks good.  I assume this accounts for the weirdness that could ensure when a topology change occurs and one of the replicas returns a VbucketMapOutdated?

brett19:
@Westwooo Perhaps this is already handled since VbucketMapOutdated is in our default retry strategy, which should probably be getting applied (if its not already) to the individual operations being performed.  Relatedly I noticed in your first comment: "we will not return any more results from that node, even if we retrieve the doc", I had assumed that once a 'replica fatal error' error occurs for a particular node, we would not continue attempting to retrieve the document from that replica.

Westwooo:
> @Westwooo Perhaps this is already handled since VbucketMapOutdated is in our default retry strategy, which should probably be getting applied (if its not already) to the individual operations being performed. Relatedly I noticed in your first comment: "we will not return any more results from that node, even if we retrieve the doc", I had assumed that once a 'replica fatal error' error occurs for a particular node, we would not continue attempting to retrieve the document from that replica.

We are using the default retry manager,. which has `ErrVbucketMapOutdated` in it's list of retry-able errors. And we continue the loop for a replica ID even if we have returned a result for that replica ID incase there is a rebalance while we are getting the replicas and the replica ID for that loop is routed to a different node, in which case we would return another result for that replica ID.

brett19:
> And we continue the loop for a replica ID
Ahh, so we do loop through the replicas again, but if they point to a server we already have an answer from we don't bother sending the actual operation on the wire.  I see.

brett19 (approved):

PR #339, opened 2025-09-15
ING-784: Include server duration in memdx errors for crud ops

chvck:


memdx/errors.go
```diff
@@ -260,3 +261,19 @@ func (e DcpRollbackError) Error() string {
...
 	return ErrDcpRollback
 }
+
+type ErrorWithServerDuration struct {
+	Cause          error
+	ServerDuration time.Duration
+}
+
+func (e ErrorWithServerDuration) Error() string {
+	return fmt.Sprintf(
```
chvck, line 271: Unsure we actually want this rather than just passthrough of the message.

Westwooo (dismissed):

memdx/errors.go
brett19: We do, most other 'enhancement' style error wrappers add their context to the Error string and it makes it a lot easier to debug when you can see what wrappers exist around an error you're printing.

memdx/ops_crud.go
```diff
@@ -155,6 +155,13 @@ func (o OpsCrud) decodeCommonError(resp *Packet) error {
 	return OpsCore{}.decodeError(resp)
 }
 
+func (o OpsCrud) wrapErrorWithServerDuration(err error, serverDuration time.Duration) *ErrorWithServerDuration {
+	return &ErrorWithServerDuration{
```
brett19, line 159: I think this will end up wrapping errors with a ErrorWithServerDuration object even in cases where the server does not support server durations, or did not pass one back, is that going to be an issue?  I wonder if the server returning "0" is valid, if not, you can probably just do something like..
```
if serverDuration == 0 {
return err
}

brett19 (dismissed):

Westwooo (approved):

PR #341, opened 2025-09-25
ING-1295: Remove manifestID from OrchestrateMemdCollectionID cb

chvck:


crud_test.go
```diff
@@ -62,21 +62,14 @@ func TestSimpleCrudCollectionMapOutdatedRetries(t *testing.T) {
...
-	)
+	res, err := OrchestrateSimpleCrud[*UpsertResult](context.Background(), rs, cr, vb, nil, nkcp, "scope", "collection", 0, []byte("somekey"), fn)
 	require.NoError(t, err)
 	require.NotNil(t, res)
 
 	assert.Equal(t, 2, collectionCalls)
 	assert.Equal(t, 2, fnCalls)
 }
+
+func TestSimpleCrudCollectionIDOnlyReturnsImmediately(t *testing.T) {
```
brett19, line 73: :O

brett19 (changes requested):

vbucketconsistency.go
```diff
@@ -0,0 +1,82 @@
...
+				RequestedVbId: vbID,
+				RequestVbUUID: vbUUID,
+				ActualVbUUID:  serverVbUUID,
+			}
+		}
+
+		return nil
+	}
+
+	vb.slowLock.Lock()
```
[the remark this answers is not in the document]
chvck: It prevents us from writing  2 different caches into the baggage under the same key, which could lead to racey behaviour where 2 calls to Baggage(...) return different caches. Possibly unimportant but seems edge casey and hard to debug.

crud_test.go
chvck: Are you questioning the quality of my excellent test? It's guaranteed to pass!

brett19: My mistake. +2

PR #389, opened 2026-07-03
ING-2050: Ensure user passwords

Westwooo:
## Summary

Adds the ability for `EnsureUser` to confirm a password change has propagated to all nodes, using the server's `password_change_date` field.

### Design notes

- `UpsertUser` now internally reads the user's current `password_change_date` *before* issuing the PUT, and returns it as `UpsertUserResult.PreviousPasswordChanged`. 
- **The marker is an opaque type (`PasswordChangedMarker`)**, not a raw `time.Time`. It can only be constructed from a `GetUser`/`UpsertUser` response, so callers can't accidentally pass their own wall-clock time into `EnsureUser` (which would be unsafe due to potential clock skew and gives no real guarantee).
- **The marker is only populated/checked when the password actually changes.** `UpsertUser` returns `nil` when `opts.Password == ""`, and `EnsureUser`'s `SincePasswordChanged` is optional — otherwise a roles/groups-only update would make `EnsureUser` wait forever on a timestamp that will never move.
- **New user creation uses a zero-value baseline.** If the pre-write probe returns "user not found," that's treated as a valid baseline (rather than an error) — any real `password_change_date` set during creation will always be "after" it.
- Added a dino-based integration test (`TestEnsureUserDino`) mirroring the existing `EnsureBucket`/`EnsureManifest` dino tests, since `EnsureUser` previously had no test coverage of any kind.
- **Verified that resubmitting an identical password still bumps `password_change_date`** (confirmed against a real cluster) — the server doesn't diff against the existing password hash, so this can't cause `EnsureUser` to hang waiting on a timestamp that never changes.


Westwooo:
@droid review

cbmgmtx/ensureuserhelper.go
```diff
@@ -64,10 +90,85 @@ func (e *EnsureUserHelper) pollOne(
...
+			}
+		}
+		if !isDirectlyAssigned {
+			continue
+		}
+
+		actualUserRoles = append(actualUserRoles, encodeRole(role.RoleJson))
+	}
+
+	return stringSetsEqual(actualUserRoles, want)
```
chvck: It seems odd that we're string matching these? Would we be better translating the RoleJson and RoleWithOriginsJson into a proper type?

Westwooo: Good point, if we got a malformed role then the string match would fail and we'd poll forever. I changed it so that we decode the string array into the appropriate JSON and compare that.

PR #391, opened 2026-07-20
ING-2006: Stop logging Authenticator password in config log lines

safianalicb:
`gocbcorex` logged the `Authenticator` — which can contain the user's plaintext password — when logging config structs via `zap.Any`, including at `info` level in the kv client babysitter.

The fix redacts the authenticator wherever these configs are logged, and hardens the secret-holder itself:

- Config types (`kvClientBabysitterClientConfig`, the bucket/node watcher reconfigure configs) now implement `MarshalLogObject` that omits the auth data, with their log calls switched to `zap.Object`.
- `PasswordAuthenticator` implements `MarshalJSON` and `MarshalLogObject` so any reflection-based logging that reaches it redacts the password rather than printing it.
- The cbauth revrpc unknown-method handler no longer logs raw params, which may carry credentials.

Note: the `MarshalJSON` change affects JSON serialization of `PasswordAuthenticator` everywhere, not just logging. No consumer in this repo round-trips it through JSON.

### DIFF
--- a/bucketswatcher_http.go
+++ b/bucketswatcher_http.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/exp/slices"
 
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 type BucketsWatcherHttpConfig struct {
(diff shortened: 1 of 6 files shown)


bucketswatcher_http.go
```diff
@@ -33,6 +34,17 @@ type BucketsWatcherHttpReconfigureConfig struct {
 	Authenticator    Authenticator
 }
 
+// MarshalLogObject implements zapcore.ObjectMarshaler, deliberately omitting
+// the authenticator which can contain the user's password.
+func (c BucketsWatcherHttpReconfigureConfig) MarshalLogObject(enc zapcore.ObjectEncoder) error {
```
Westwooo, line 39: In these comments where we omit the authenticator is that the same password authenticator that you added custom marshaling for above? If so can we not leave the marshalling of the things that include the authenticator as is and they will fall back to our custom authenticator marshalling that excludes the password? Or does it not work like that?

safianalicb: > If so can we not leave the marshalling of the things that include the authenticator as is and they will fall back to our custom authenticator marshalling that excludes the password?

We could, but since `Authenticator` is an interface, if new implementations ones are added in the future, then they will be logged (if `MarshalJSON`/`MarshalLogObject` aren't implemented). This precaution means that confidential information won't be leaked, regardless of what `Authenticator` is used.

Westwooo (approved):

